Issue #21, opened 2014-03-20
/root filesystem performance improvements


pleasantone:
Some research was done here[1] and adapted by the OpenElec raspberry pi folks for significantly increasing the SD card performance by block aligning the ext4 filesystem such that read/write/erases drop on sector boundaries.

While determining the actual block alignments (4k/8k/16k) might be a bit of a pain, we win at all three alignments if we start the root filesystem on a block that is a modulo of 32768 (16k at 512 bytes/block).

Then people can easily fine-tune ext4 sector sizes if you allow us access to the mkfs.ext4 options (e.g. someone with 4k alignment might do: -O ^has_journal -E stride=2,stripe-width=1024 -b 4096)

Thoughts?

[1] http://blogofterje.wordpress.com/2012/01/14/optimizing-fs-on-sd-card/

diederikdehaas:
The options are available and I have no intention of changing the defaults, so it's time to close this issue.

PR #152, opened 2014-10-25
Implement SD enhancements, mostly for ext4

steveschnepp:
This is a PR to follow guidelines found in #21.

It is not intended for direct merging, but more for discussing issues around it.


diederikdehaas:
While it was an interesting discussion, I see no need to keep this PR open (as it won't be merged).

steveschnepp:
true. Thanks for the discussion :+1:

PR #172, opened 2014-12-27
Do not create a new partition table if the drive is already partitioned

socram8888:
Basically I manually specified a partition, hoping that the installer won't touch the other partitions which had valuable data, just to discover later that the MBR was wiped :v: 

This PR modifies the installation script so that if the partition is manually specified the drive isn't repartitioned, because we already have a partition ready to be used.

This modification also raises another question: why repartition the SD if the root partition is in an external drive? Can't we just format the boot partition and leave the MBR and the rest of the partitions untoched? Either way I consider this out of the scope of this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@
 /raspbian-ua-netinst-*.img
 /raspbian-ua-netinst-*.img.xz
 /raspbian-ua-netinst-*.img.bz2
+installer-config.txt
+
+# Ignore backup files created by gedit
+*~
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -335,7 +335,10 @@ if [ "$usbroot" != "" ]; then
     echo "OK"
 fi
 
+manually_specified_root_partition=true
 if [ "$rootpartition" = "" ]; then
+    manually_specified_root_partition=false
+
     if [ "$rootdev" = "/dev/sda" ]; then
         rootpartition=/dev/sda1
     else
@@ -418,9 +421,13 @@ w
 EOF
     echo "OK"
 
-    echo -n "Applying new partition table for $rootdev... "
-    dd if=/dev/zero of=$rootdev bs=512 count=1 &>/dev/null
-    fdisk $rootdev &>/dev/null <<EOF
+    # If the user manually specified a partition, we can suppose the drive is
+    # already partitioned and therefore we don't need to create a new
+    # partition table
+    if [ "$manually_specified_root_partition" = "false" ]; then
+        echo -n "Applying new partition table for $rootdev... "
+        dd if=/dev/zero of=$rootdev bs=512 count=1 &>/dev/null
+        fdisk $rootdev &>/dev/null <<EOF
 n
 p
 1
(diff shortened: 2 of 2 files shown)


ShorTie8:
this is untested
but i think it will work for ya

--- rcS.orig	2014-12-31 19:07:39.229445184 -0500
+++ rcS	2015-01-01 07:35:20.000000000 -0500
@@ -384,6 +384,37 @@
     echo -n "$i.. "
 done
 
+# these shouldn't really be changed unless auto-detection fails
+#bootdev=/dev/mmcblk0
+#bootpartition=/dev/mmcblk0p1
+#rootdev=/dev/mmcblk0
+#rootpartition=
+
+THIRD_PART=$(fdisk -l /dev/mmcblk0 | grep 'p3' |  awk '{print $2}')
+if [ "$THIRD_PART" != ""  ]; then
+
+    echo -n "Applying new partition table... "
+    rootsize=$(fdisk -l /dev/mmcblk0 | grep 'p2' |  awk '{print $3}')
+    dd if=/dev/zero of=$bootdev bs=512 count=1 &>/dev/null
+    fdisk $bootdev &>/dev/null <<EOF
+n
+p
+1
+
+$bootsize
+t
+b
+n
+p
+2
+
+$rootsize
+w
+EOF
+    echo "OK"
+
+else
+
 if [ "$rootdev" = "$bootdev" ]; then
     echo -n "Applying new partition table... "
     dd if=/dev/zero of=$bootdev bs=512 count=1 &>/dev/null
@@ -432,6 +463,10 @@
 
 fi
 
+
+fi
+
+
 echo -n "Initializing /boot as vfat... "
 mkfs.vfat $bootpartition &>/dev/null || fail
 echo "OK"

ShorTie8:
Sorry, that was my 1st post and didn't turn out real well .. :/~
here is the real diff
https://www.dropbox.com/s/5j9n8vsfnd85urt/third_part.diff

diederikdehaas:
I've added \`\`\` on a line by itself above and below your code block, so it will be seen as code as not as markdown

kpfleming:
It would be more effective to post this as a Gist here on GitHub, or even to fork the repository, put the changes into the fork and then send a pull request. Then GitHub will display the code properly, and it can be reviewed and merged if suitable.

ShorTie8:
Sorry, i just joined to add my 2 cents from things i've been playing with.
i threw it together in about 5 minutes.
if it works, it could be cleaned up a bit and better formated

Have A Great Day
ShorTie

diederikdehaas:
It's much appreciated @ShorTie8 :+1:

goranche:
> Sorry, i just joined to add my 2 cents from things i've been playing with

@ShorTie8 no need to apologise, and you'll get the hang of github sooner or later :+1: I'm sure @kpfleming didn't mean it in a bad way, but as pointers, so you'll get better with github ;)
(we've all been through the learning phase :blush: )

diederikdehaas:
The code as is won't be merged, thus I'm closing this PR.

I'm investigating another way to partitioning, so it _may_ be doable what you want, just in another form.

Issue #52, opened 2014-05-11
Additional drivers support (was: USB ethernet dongle support)


shezzski:
I have a PI with a unusable Ethernet port and are forced to use a USB Ethernet adapter, which is an ASIX AX88178, which works fine normally in Raspbian.  Unfortunately the device does not work with this installer and is unable to obtain an IP via DHCP.  The PI seems to recognise the device in dmesg.  Is it possible to get this to work?

diederikdehaas:
I'm 'merging' issue #319 with this one as it's essentially the same thing.

Thus it's probably a good idea to also include drivers from the 'staging' section, like rtl8188eu, of the kernel source.

Mausy5043:
Is that solving #44?

diederikdehaas:
No, it will only expand the drivers available for the installer kernel a bit, but there's more to it. 

The main problem is that (wireless) drivers are tightly coupled with the kernel version. Since we control the kernel version used during installation, that's not really an issue and including staging drivers should be relatively straightforward. 

But the issue which still needs a proper solution is having wireless drivers for the installed system. For drivers which are part of the kernel source/tree it's still not a problem since they're updated for every release. This means that even if the kernel on the installed system is say 4.3 it should still have working drivers. 

But drivers which are not part of the kernel source, which unfortunately are a lot, that becomes an issue. If the kernel version of the installer and the installed system are the same, we could go for a shortcut and just copy those drivers from the installer to the installed system.
But things will fail when the kernel versions are different between installer and installed system.
What you then actually need are the kernel sources and compile the relevant drivers (or all) on the spot, preferably with [DKMS](http://linux.dell.com/dkms/).

Mausy5043:
Please note that PR #394 (on the v1.1.x branch) now allows users to inject their "own" drivers/modules into the installed system.

Mausy5043:
Please note that PR #397 provides WiFi support to the installer. Would love to hear from @shezzski if this helps.

Mausy5043:
Closing issue, because no response in 4 months. If you feel this problem is not resolved please feel free to re-open and add a comment providing additional information.

Issue #44, opened 2014-04-15
Installing over wireless?


nodiscc:

JedMeister:
This is a very cool thing. IMO you should also document which chips it supports OOTB and how to include other ones...

goranche:
> What wlan drivers should be included in the installation? So far I have only included realtek because that's the one I use.

I was about to say "all the [working ones](http://elinux.org/RPi_USB_Wi-Fi_Adapters)", but damn, that's a large list... :smile: 

I'd say PR it, and maybe include some info on how to add additional drivers, and let others add the drivers they need... :+1:

diederikdehaas:
> > What wlan drivers should be included in the installation? So far I have only included realtek because that's the one I use.

> I was about to say "all the working ones", but damn, that's a large list...

And then it's still incomplete :wink:  as I have an [Edimax EW-7822UAC (AC1200)](http://www.edimax.com/edimax/merchandise/merchandise_detail/data/edimax/global/wireless_adapters_ac1200_dual-band/ew-7822uac) with a [working driver](https://github.com/diederikdehaas/rtl8812AU) :stuck_out_tongue_winking_eye: 

Don't worry about the drivers for now ... that'll come. There actually [already is an issue for that](https://github.com/debian-pi/raspbian-ua-netinst/issues/52)

goranche:
> And then it's still incomplete

that's why I say "PR it" and let the people with additional hardware do the boring work of adding drivers :angel: :wink:

JedMeister:
> let the people with additional hardware do the boring work of adding drivers

:+1:

diederikdehaas:
> let the people with additional hardware do the boring work of adding drivers

We could add the in-tree kernel wireless drivers to the installation image, which would add 1.4MB (tar.gz compressed) to the installation image. Motivation for this is that they would (already) be available in the installed system.
For out-of-tree kernel drivers I agree with you to let others do the work themselves. 
(Best way would be to create `\<driver\>-dkms` debian packages so wireless won't stop working when a new kernel gets installed)

One possible solution in which we could facilitate custom drivers is by letting the user create a drivers.tar.gz archive which matches the tree structure for kernel modules (ie `/lib/modules/<kernel-version>/kernel/drivers/`)

```sh
if [ -f /bootfs/config/drivers.tar.gz ] ; then
   echo -n "Loading drivers... "
   tar -zxf /bootfs/config/drivers.tar.gz -C /
[...]

PR #175, opened 2015-01-04
Added configuration for installing over wifi

wiguy:
User must add wpa_supplicant.conf file to /boot/config/ which is in wpa_supplicant-format, see man wpa_supplicant.conf.

Added following options and added it to readme
ifname - interface name used for network connection
wlandriver - filename of driver which is inserted into the kernel. This file must exist in the default installation.

Installation now includes wpa_supplicant and firmware-realtek package.
Directory /drivers/wireless holds wlan drivers

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,11 +131,13 @@ The format of the file and the current defaults:
     timezone=Etc/UTC # set to desired timezone (e.g. Europe/Ljubljana)
     locales=  # a space delimited list of locales that will be generated during install (e.g. "en_US.UTF-8 nl_NL sl_SI.UTF-8")
     system_default_locale= # the default system locale to set (using the LANG environment variable)
+    ifname=eth0
     ip_addr=dhcp
     ip_netmask=0.0.0.0
     ip_broadcast=0.0.0.0
     ip_gateway=0.0.0.0
     ip_nameservers=
+    wlandriver=
     online_config= # URL to extra config that will be executed after installer-config.txt
     usbroot= # set to 1 to install to first USB disk
     cmdline="dwc_otg.lpm_enable=0 console=ttyAMA0,115200 kgdboc=ttyAMA0,115200 console=tty1 elevator=deadline"
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ The format of the file and the current defaults:
     ip_broadcast=0.0.0.0
     ip_gateway=0.0.0.0
     ip_nameservers=
-    wlandriver=
+    drivers_to_load=
     online_config= # URL to extra config that will be executed after installer-config.txt
     usbroot= # set to 1 to install to first USB disk
     cmdline="dwc_otg.lpm_enable=0 console=ttyAMA0,115200 kgdboc=ttyAMA0,115200 console=tty1 elevator=deadline"
--- a/build.sh
+++ b/build.sh
@@ -106,6 +106,7 @@ mkdir -p rootfs/sbin/
 mkdir -p rootfs/usr/bin/
 mkdir -p rootfs/usr/share/
 mkdir -p rootfs/usr/share/keyrings/
+mkdir -p rootfs/drivers/wireless
 
 cp -a tmp/lib/modules/${KERNEL_VERSION}/modules.{builtin,order} rootfs/lib/modules/${KERNEL_VERSION}
 
@@ -200,6 +201,7 @@ cp tmp/lib/*/libcom_err.so.2.1 rootfs/lib/libcom_err.so.2
 
 # libgcc1 components
 cp tmp/lib/*/libgcc_s.so.1 rootfs/lib/
+cp tmp/lib/*/librt.so.1 rootfs/lib/
 
 # liblzo2-2 components
 cp tmp/lib/*/liblzo2.so.2 rootfs/lib/liblzo2.so.2
@@ -209,6 +211,30 @@ cp tmp/lib/*/libuuid.so.1.3.0 rootfs/lib/libuuid.so.1
 
 # zlib1g components
 cp tmp/lib/*/libz.so.1  rootfs/lib/libz.so.1
+
+# wpa_supplicant components
+cp tmp/sbin/wpa_supplicant rootfs/sbin/wpa_supplicant
+cp -r tmp/etc/wpa_supplicant rootfs/etc/wpa_supplicant
+
+# firmware realtek
+cp tmp/lib/modules/*/kernel/drivers/net/wireless/rtl8192cu/8192cu.ko rootfs/drivers/wireless/8192cu.ko
+
+# libssl components
+cp tmp/usr/lib/*/libssl.so.1.0.0 rootfs/lib/libssl.so.1.0.0
+cp tmp/usr/lib/*/libcrypto.so.1.0.0 rootfs/lib/libcrypto.so.1.0.0
+
+# libdbus components
+cp tmp/lib/*/libdbus-1.so.3 rootfs/lib/libdbus-1.so.3
+cp tmp/lib/*/libdl.so.2 rootfs/lib/libdl.so.2
+
+# libnl components
+cp tmp/lib/*/libnl-3.so.200 rootfs/lib/libnl-3.so.200
+
+# libnl-genl components
+cp tmp/lib/*/libnl-genl-3.so.200 rootfs/lib/libnl-genl-3.so.200
+
+# libpcsclite components
+cp tmp/usr/lib/*/libpcsclite.so.1 rootfs/lib/libpcsclite.so.1
 
 cd rootfs && find . | cpio -H newc -ov > ../installer.cpio
 cd ..
--- a/build.sh
+++ b/build.sh
@@ -90,7 +90,7 @@ rm -rf bootfs
 mkdir -p bootfs
 
 # raspberrypi-bootloader-nokernel components and kernel
-cp tmp/boot/* bootfs/
+cp -r tmp/boot/* bootfs/
 rm bootfs/System*
 rm bootfs/config-*
 mv bootfs/vmlinuz* bootfs/kernel_install.img
@@ -106,7 +106,6 @@ mkdir -p rootfs/sbin/
 mkdir -p rootfs/usr/bin/
 mkdir -p rootfs/usr/share/
 mkdir -p rootfs/usr/share/keyrings/
-mkdir -p rootfs/drivers/wireless
 
 cp -a tmp/lib/modules/${KERNEL_VERSION}/modules.{builtin,order} rootfs/lib/modules/${KERNEL_VERSION}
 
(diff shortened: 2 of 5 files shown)


update.sh
```diff
@@ -17,6 +17,8 @@ packages="$packages e2fsprogs"
 packages="$packages f2fs-tools"
 packages="$packages gpgv"
 packages="$packages raspbian-archive-keyring"
+packages="$packages wpasupplicant"
+packages="$packages firmware-realtek"
```
diederikdehaas: Could you order them alphabetically? 
It's probably also a good idea to also include `firmware-linux-free`, `firmware-linux-nonfree` and probably various others.

scripts/etc/init.d/rcS
```diff
@@ -788,6 +813,18 @@ else
     echo "FAILED !"
 fi
 
+# if eth0 was not used during initial boot, a wireless interfaces was used
+if [ "$ifname" != "eth0" ]; then
+    # Install wireless dependencies
+    echo -n "Installing wpasupplicant package... "
+    chroot /rootfs /usr/bin/apt-get -y install wpasupplicant firmware-realtek >& /dev/null
```
diederikdehaas: I agree with `wpasupplicant`, but I'm leaning towards letting the firmware package be installed by the user (through configuration)

scripts/etc/init.d/rcS
```diff
@@ -217,19 +232,26 @@ for i in `seq 1 10`; do
     echo -n "$i.. "
 done
 
+if [ "$ifname" != "eth0" ]; then
+    # if eth0 is not the chosen eth interface it is a wireless interface and wpa_supplicant must connect to wlan
+    echo -n "Running wpa_supplicant... "
+    wpa_supplicant -B -Dwext -c/bootfs/config/wpa_supplicant.conf -iwlan0 >& /dev/null || fail
```
diederikdehaas: I'd change `-Dwext` to `-Dnl80211,wext`. `nl80211` is the 'new' wireless standard so I'd very much like to support that. `wext` is deprecated, but without it a lot of wireless cards probably won't work.
The `-iwlan0` should probably be `-i$ifname`

scripts/etc/init.d/rcS
```diff
@@ -217,19 +232,26 @@ for i in `seq 1 10`; do
...
+    [ $? -eq 0 ] && echo "OK"
```
diederikdehaas: You can do a simple `echo "OK"` here since `|| fail` already means a return value != 0 will fail it.
Besides, if you check the return value, you should also handle the non-zero case.

scripts/etc/init.d/rcS
```diff
@@ -198,11 +204,20 @@ fi
 echo "  online_config = $online_config"
 echo
 
-echo -n "Waiting for eth0... "
+echo "Waiting for $ifname... "
+
+if [ "$ifname" != "eth0" ]; then
+    # if eth0 is not the chosen eth interface it is a wireless interface and a driver must be loaded
+    echo -n "Loding driver $wlandriver... "
```
diederikdehaas: Loding -> Loading

scripts/etc/init.d/rcS
```diff
@@ -198,11 +204,20 @@ fi
...
+    insmod /drivers/wireless/$wlandriver || fail
```
diederikdehaas: Why `insmod` and not `modprobe`?

build.sh
```diff
@@ -210,6 +212,30 @@ cp tmp/lib/*/libuuid.so.1.3.0 rootfs/lib/libuuid.so.1
...
+
+# libnl components
+cp tmp/lib/*/libnl-3.so.200 rootfs/lib/libnl-3.so.200
+
+# libnl-genl components
+cp tmp/lib/*/libnl-genl-3.so.200 rootfs/lib/libnl-genl-3.so.200
+
+# libpcsclite components
+cp tmp/usr/lib/*/libpcsclite.so.1 rootfs/lib/libpcsclite.so.1
+
```
diederikdehaas: Also order these alphabetically, with the split between programs and libraries, just like in `update.sh`

diederikdehaas:
First of all: nice work :+1: 
I'll add various remarks in the code itself ...

build.sh
```diff
@@ -210,6 +212,30 @@ cp tmp/lib/*/libuuid.so.1.3.0 rootfs/lib/libuuid.so.1
 # zlib1g components
 cp tmp/lib/*/libz.so.1  rootfs/lib/libz.so.1
 
+# wpa_supplicant components
+cp tmp/sbin/wpa_supplicant rootfs/sbin/wpa_supplicant
+cp -r tmp/etc/wpa_supplicant rootfs/etc/wpa_supplicant
+
+# firmware realtek
```
diederikdehaas: This is not the firmware, but the driver/module :wink:

build.sh
```diff
@@ -210,6 +212,30 @@ cp tmp/lib/*/libuuid.so.1.3.0 rootfs/lib/libuuid.so.1
...
+cp tmp/lib/modules/*/kernel/drivers/net/wireless/rtl8192cu/8192cu.ko rootfs/drivers/wireless/8192cu.ko
```
diederikdehaas: Why place it in `rootfs/drivers/wireless/` and not in `/lib/modules/$(uname -r)/kernel/drivers/net/wireless/rtl8192cu/`?

README.md
```diff
@@ -131,11 +131,13 @@ The format of the file and the current defaults:
     timezone=Etc/UTC # set to desired timezone (e.g. Europe/Ljubljana)
     locales=  # a space delimited list of locales that will be generated during install (e.g. "en_US.UTF-8 nl_NL sl_SI.UTF-8")
     system_default_locale= # the default system locale to set (using the LANG environment variable)
+    ifname=eth0
     ip_addr=dhcp
     ip_netmask=0.0.0.0
     ip_broadcast=0.0.0.0
     ip_gateway=0.0.0.0
     ip_nameservers=
+    wlandriver=
```
diederikdehaas: Just a thought: how about changing it to `drivers_to_load` (or sth like that) and make that a space seperated list?
That way we can use that to load various drivers if needed and it is also more generic. That way we could also support [USB ethernet dongles](https://github.com/debian-pi/raspbian-ua-netinst/issues/52) or things like that.
(it may cause an issue with the selector `$ifname != eth0` though)

update.sh
wiguy: Will do. However, I'll ignore the "various others" unless you have some specific list.
What to include here is part of the "what to do with drivers"-discussion.

diederikdehaas: On Thursday 08 January 2015 09:29:17 wiguy wrote:
> unless you have some specific list.

What do you mean?

> What to include here is part of the "what to do with drivers"-discussion.

Added my input to issue #44

build.sh
wiguy: This cleared up a misunderstanding on my part. I'm just learning this stuff on the way :)
I'm not even using the firmware-realtek package. I just ran through the installation without it. Why add more firmware packages? (ref first comment on update.sh)

Somethings come up, so next commit will have to wait for some more time.

diederikdehaas: > Why add more firmware packages?

Because various/a lot of wireless network cards need firmware to work.

wiguy:
Hi again. I've fixed most of your remarks. With one exception.
I do not yet handle firmware. It takes quite a lot of space, and does not fit in the alotted 32mb.
I'm sure it is possible to expand the partition... But I haven't tried very hard yet. 

Another option is to have the user provide firmware.

Right now nothing is included.
I've googled a little bit and the following list seems reasonable to include:
packages="$packages firmware-atheros"
packages="$packages firmware-brcm80211"
packages="$packages firmware-ipw2x00"
packages="$packages firmware-iwlwifi"
packages="$packages firmware-libertas"
packages="$packages firmware-linux-free"
packages="$packages firmware-linux-nonfree"
packages="$packages firmware-ralink"
packages="$packages firmware-realtek"

diederikdehaas:
Hi! Thanks for your improvements :+1: 

With the $drivers_to_load (space separated list) I actually had something like this in mind:
```sh
if [ "$drivers_to_load" != "" ] ; then
   echo "Loading additional drivers."
   for driver in $drivers_to_load
   do
      echo -n " Loading driver '$driver' ... "
      modprobe $driver
      if [ $? -eq 0 ] ; then
         echo "OK"
      else
         echo "FAILED !"
      fi
   done
   echo "Finished loading additional drivers"
fi
```
This way if there's a need for multiple drivers to be loaded it would work too.

>  does not fit in the allotted 32mb

You can expand that by increasing the [`count` value in buildroot.sh](https://github.com/debian-pi/raspbian-ua-netinst/blob/v1.1.x/buildroot.sh#L11).

I can merge your changes right now and have further improvements by you/me/others done through separate PRs or we can continue improving this PR till it's 'done' and then merge it.
What do you prefer?

wiguy:
I've shamelessly taken your code and pasted it.
There's a small thing to remember. When doing it like you suggested, the user should specify the driver paths using ie /lib/modules/$(uname -r)/kernel/etcetc or /lib/modules/*/kernel/etcetc.
Either that or the user must know the exact kernel version.

I think you merge my changes as is. The basic is there now. If you have an adapter not requiring any extra drivers it will work. Extending with additional drivers can be done by separate PRs.

diederikdehaas:
Thank you very much :+1:

malsony:
Thank you guys! But how to download version 1.1.x for testing? And can I name several supported wireless drivers in "drivers_to_load"? Thank you!

diederikdehaas:
There is no ready made installer for v1.1.x so you'd have to build that yourself by cloning the repository and following the instructions from [BUILD.md](https://github.com/debian-pi/raspbian-ua-netinst/blob/master/BUILD.md).
`drivers_to_load` is a space separated list, so yes, you can name several drivers.

malsony:
Got it, thanks! Already got a copy and will try it following your guide.

moritzmhmk:
I did get 1.1.x to work with my Edimax EW-7811Un but it will stop working while downloading the packages :-/

I then tried a "hama" and a "TP-Link" dongle - but those have problems with the drivers (``rt73usb`` and ``ath9k``) which depend on ``mac80211`` and ``cfg80211`` (the error I get is "...unknown symbol in module..." - but I found that this happens with missing dependencies and modinfo shows those two)

How can I add ``mac80211`` and ``cfg80211``?

diederikdehaas:
> How can I add mac80211 and cfg80211?

Those are the wireless extensions from the linux kernel. The driver(s) need to be not ancient for it to support those. A quick look indicated that `rt73usb` and `atk9k` are part of the linux kernel, so they should support it.
It's been a _long_ time since I tried this myself, so I would have to dive in again to see if and how well it (still) works. On the plus side I do have the Edimax EW-7811Un myself. On the down side is the lack of time, so I don't know when I'll be able to get to it.

moritzmhmk:
I did use ``jessie``

I found another wifi dongle that has a ``RTL8192`` chipset (while the Edimax has ``RTL8188``) but uses the same ``rtl8192cu`` driver. It did work just fine :)

Still it would be nice to know where those errors with the ``ath9k_htc`` (I forgot the htc before) and ``rt73usb`` come from. I can modprobe those on the finished system - but not before the installation.

diederikdehaas:
Did you install a firmware package? I _think_ that the `ath9kX` drivers need those, while others may work without them.
There's a reasonable chance that some serious work needs to be done before the installer works properly over wireless for most devices (all is pretty much not possible). 
If you'd like to help testing that when extra code is written, that would be great.

Issue #80, opened 2014-08-10
No DNS lookup in built image


goranche:
**Note**: this issue has been "resolved" by a workaround, which will stay in place until a permanent solution is ready and tested. busybox-static (the cause of these problems) will be replaced with busybox, and all our troubles will seem so far away :+1: 
:wink: 

=========================================

first off, thanks for this project, love it...
so much so, that I started coding some additional features (starting with locale and timezone from #35), but I'm experiencing some really strange issues when building the install image myself...

first, when running ./build.sh, I'm getting depmod warnings:
depmod: WARNING: could not open /home/goranb/raspbian-ua-netinst/rootfs/lib/modules/3.10-3-rpi/modules.order: No such file or directory
depmod: WARNING: could not open /home/goranb/raspbian-ua-netinst/rootfs/lib/modules/3.10-3-rpi/modules.builtin: No such file or directory

second, when booting from the resulting image, the system doesn't seem to be able to do DNS lookupups.
it get's an IP address, ping works fine, there is a nameserver entry in /etc/resolv.conf (the right one for my network), so ping 74.125.232.16 works, but ping www.google.com doesn't

this doesn't happen when booting the release image, but when I replace installer.cpio (to test my code), the problem starts...

to rule out any problem with my code, I've tried generating an image from the code in the repository (so without my changes), same thing...

any ideas?

the system I'm building on is a 64 bit ubuntu, but I've tried it on raspbian (default from raspberrypi.org) as well...

diederikdehaas:
The current version of raspbian jessie busybox-static does not have this problem, thus closing.

PR #183, opened 2015-02-24
rcS: add offset for boot/root partitions

steveschnepp:
The number is expressed in "blocks". That means a 512 byte unit.
Therefore aligning on 64MiB means "131072".

The partitions usually needs to be aligned, specially on SD cards. This
enables a basic alignment.

Fully automated alignement isn't done because the /boot patition is
usually of a fixed sized, and the offset should be the same for SD size.

Note that the default might be set to something like 131072, which
corresponds nicely to the default 50M /boot.

This PR is loosely related to #152, but not dependent.

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -11,7 +11,9 @@ hostname=pi
 domainname=
 rootpw=raspbian
 cdebootstrap_cmdline=
+bootoffset=
 bootsize=+50M
+rootoffset=
 rootsize=
 timeserver=time.nist.gov
 ip_addr=dhcp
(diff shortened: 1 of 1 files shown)


steveschnepp:
Hmmm... actually 770b40198e4e9a2b8f670382707d410e3946a94b from #152 is much better. As it is not bugged (missed a "u"), and it has a 64MiB default

diederikdehaas:
As part of the solution to https://github.com/debian-pi/raspbian-ua-netinst/issues/80 busybox's fdisk will be replaced by the normal fdisk (which does alignment correctly by default).
Would this PR (and PR #152) still make sense then?

steveschnepp:
I'd say yes, as the alignements aren't the same on every SD. That said the extra "u" might not be required, and this PR has to be reworked, but not dropped IMHO.

diederikdehaas:
Ok.
I'm not in favor of changing (fdisk's) defaults though.

diederikdehaas:
I am closing this PR as it should be targetted to the v1.1.x branch.

I'm certainly not ruling out that this feature will be implemented, but it will be in v1.1.x (or later) and maybe even in a different form.

Issue #185, opened 2015-03-03
Syntax error when running update.sh


ChrisMagellan:
When I'm running update.sh I'm getting,

update.sh: 61: update.sh Syntax error: "(" unexpected (expecting "}")

Any ideas? I confirmed the required packages were installed. I'm running Ubuntu 64-bit Server 14.10.

goranche:
what version are you using, or did you use the "latest and greatest" from github?

ChrisMagellan:
1.0.6

goranche:
@ChrisMagellan could you also check your bash version, please :thought_balloon:

ChrisMagellan:
GNU bash, version 4.3.30(1)-release (x86_64-pc-linux-gnu)

goranche:
:smile: ok, even newer than mine... weird

ChrisMagellan:
Yeah, it's a pretty new install of Ubuntu Server.

goranche:
@ChrisMagellan could you try running the script directly using bash:

    bash update.sh

the line you're seeing the error in creates an array, which is a bash extension (that's why I asked about the version)... I have a suspicion that your bash isn't really bash (not sure how that would come to be, though)

ChrisMagellan:
Ah, yep, that's working now. Thanks!

goranche:
@diederikdehaas you think might want to change the shebang lines to something like this:

    #!/usr/bin/env bash

also, reopening this issue, as it might lead to an improvement of the scripts :wink:

goranche:
> yep, that's working now

@ChrisMagellan you're welcome... but I'm still reopening the issue, because this is something that could probably be handled better in the scripts... thanks for the issue :+1:

ChrisMagellan:
Thank you!

diederikdehaas:
It's working fine on my  (Debian Sid) machine and my bash version is 
`GNU bash, version 4.3.33(1)-release (x86_64-pc-linux-gnu)`

@ChrisMagellan What does `ls -l /bin/bash` output with you? I get
`-rwxr-xr-x 1 root root 1033720 Jan 28 17:59 /bin/bash`

ChrisMagellan:
`-rwxr-xr-x 1 root root 1029720 Oct  7 11:15 /bin/bash`

diederikdehaas:
Thanks Chris, I see no important difference there.
[...]

PR #187, opened 2015-03-03
Changed shebang lines to use env

goranche:
this will hopefully make sure bash is used to run the build scripts


goranche:
this addresses #185 , but for the other build scripts as well

PR #188, opened 2015-03-04
Updated readme.md to make more clear

damienrj:
It was unclear to me that you could place the installer-config.txt on the sdcard without building your own install.  I tried to make it more clear. I think it just needs to be mentioned explicitly in the readme file.


goranche:
sounds like a reasonable change... :smile:

Issue #190, opened 2015-03-07
No space left on device (1.0.7-beta)


Mausy5043:
As promised.
Tried to build a Raspberry Pi with the 1.0.7-beta:
```
gzip: 
stdout: No space left on device
update-initramfs: failed for /boot/initrd.img-3.18.0-trunk-rpi with 1.
dpkg: error processing initramfs-tools (--configure):
 subprocess installed post-installation script returned error exit status 1
Errors were encountered while processing:
 initramfs-tools
E
: 
Sub-process /usr/bin/dpkg returned an error code (1)

E: Internal error: install

Oh noes, something went wrong!
You have 10 seconds to hit ENTER to get a shell...
```
(I could've sworn there was already an issue logged for this, but I can't find it anymore.)

Rebooting as I type with `bootsize=+64M` in `...\init.d\rcS`

diederikdehaas:
A bit more info is needed here.
What were you trying to do? When did it fail? Do you have a log file you could share? What configuration files did you use?

tais:
I have had the same issue, raspi B2, following config

preset=server
mirror=http://mirrordirector.raspbian.org/raspbian/
release=jessie
hostname=xxxxx
domainname=xxxx.local
rootpw=xxxxxxxxx
cdebootstrap_cmdline=
timeserver=nl.pool.ntp.org
ip_addr=xxx.xxx.xxx.xxx
ip_netmask=xxx.xxx.xxx.xxx
ip_gateway=xxx.xxx.xxx.xxx
ip_nameservers=xxx.xxx.xxx.xxx

Mausy5043:
Relevant part of the logfile is up there. I'm afraid I destroyed the rest. If you really need it I could re-create  another one later. But I don't see the point right now.
Some background: R-PI B-type, 4GB SD card. I built the SD-card image as usual. Then inserted it into the Pi.

installer-config.txt:
```
packages="apt-utils,avahi-daemon,bc,git,htop,libnss-mdns,logrotate,lsof,nano,nfs-common"
packages="$packages,python,raspi-copies-and-fills,rsync,screen,sudo"
hostname=raspberrypi
cmdline="dwc_otg.lpm_enable=0 console=tty1 noatime loglevel=2 cgroup_enable=memory elevator=noop"
rootfstype=f2fs
label=$hostname
```
I'm watching a new boot right now with the `bootsize=+64M`. That seems to have done the trick.

diederikdehaas:
[...]

PR #191, opened 2015-03-07
Increased bootsize

Mausy5043:
`No space left on device` during installation.
Increased bootsize from 50 to 64MB


Mausy5043:
Should fix #190

PR #203, opened 2015-03-22
Update README.md

dragetd:
Added info about the hardware random number generator. An idle RPi usually generates very few bytes of entropy per minute which can cause high delays for crypto-intensive applications. With the hardware random number generator, /dev/random manages over 25kB/s.


goranche:
nice, thanks... maybe it would be nice to start a new document, something in the line of "Improvements" or "Ok, Installed, Now what?" :thought_balloon: 

@diederikdehaas any thoughts on this?
(the only reason I'm not merging is this discussion, otherwise I'm fine with it)

diederikdehaas:
I'd say merge it :+1:

We probably should reorganize the documentation, but outside of this PR

PR #209, opened 2015-03-30
Temporary hack to fix RPi1 kernel

eLvErDe:
Force 3.12 kernel until 3.18 one is fixed on RPi1

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-KERNEL_VERSION_RPI1=3.18.0-trunk-rpi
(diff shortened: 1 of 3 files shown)


Mausy5043:
:+1: looks great. Maybe pull this (also) into the v1.0x branch?

eLvErDe:
Sure, hacky fix but I'm able to generate valid zip archives from current master. Untested on wheezy and/or rpi2 but I see no reason for this to fail.

diederikdehaas:
@plugwash will release a new kernel later today which will solve the issue pretty much.
Therefor this hack won't be needed. Thanks for the gesture though :+1:

PR #215, opened 2015-04-29
Correcting variable replacement

hleitzell:
This pull corrects these issues:

```
Configuring apt:
  Copying raspbian repo to sources.list... OK
  Replacing __RELEASE__ with jessie... OK
  Adding raspberrypi.org gpg key to apt-key.
  Copying raspberrypi.org.list to /etc/apt/sources.list.d/... sh: -eq: argument expected
FAILED
  Copying 01vcgencmd.pref to /etc/apt/preferences.d/... OK
  Replacing __RELEASE__ with jessie in 01vcgencmd.pref... grep: /rootfs/etc/apt/sources.list.d/sources.list: No such file or directory
OK
  Adding key *.key to apt.
cat: can't open '*.key': No such file or directory
gpg: no valid OpenPGP data found.
```

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -53,7 +53,7 @@ fail()
         mount $bootpartition /boot
         fail_boot_mounted=true
     fi
-    cp $LOGFILE /boot/raspbian-ua-netinst-`date +%Y%m%dT%H%M%S`.log
+    cp -- $LOGFILE /boot/raspbian-ua-netinst-`date +%Y%m%dT%H%M%S`.log
     sync
     
     # if we mounted /boot in the fail command, unmount it.
@@ -146,7 +146,7 @@ echo -n "Copying boot files... "
 # copy boot data to safety
 mount $bootpartition /boot || fail
 
-cp -r /boot/* /bootfs/ || fail
+cp -r -- /boot/* /bootfs/ || fail
 
 umount /boot || fail
 echo "OK"
@@ -519,7 +519,7 @@ mount $bootpartition /rootfs/boot || fail
 echo "OK"
 
 echo -n "Copying /boot files in... "
-cp -r /bootfs/* /rootfs/boot || fail
+cp -r -- /bootfs/* /rootfs/boot || fail
 sync
 echo "OK"
 
@@ -743,9 +743,8 @@ for listfile in $(ls *.list)
 do
     if [ $listfile != "sources.list" ] ; then
         echo -n "  Copying $listfile to /etc/apt/sources.list.d/... "
-        cp $listfile /rootfs/etc/apt/sources.list.d/
-        # verify the file is indeed present in its target location
-        if [ -f /rootfs/etc/apt/sources.list.d/$listfile ] ; then
+        cp -- $listfile /rootfs/etc/apt/sources.list.d/
+        if [ $? -eq 0 ] ; then
             echo "OK"
         else
             echo "FAILED"
@@ -768,9 +767,8 @@ done
 for preffile in $(ls *.pref)
 do
     echo -n "  Copying $preffile to /etc/apt/preferences.d/... "
-    cp $preffile /rootfs/etc/apt/preferences.d/
-    # verify the file is indeed present in its target location
-    if [ -f /rootfs/etc/apt/preferences.d/$listfile ] ; then
+    cp -- $preffile /rootfs/etc/apt/preferences.d/
+    if [ $? -eq 0 ] ; then
         echo "OK"
     else
         echo "FAILED"
@@ -863,7 +861,7 @@ fi
 # make sure there is a /boot/kernel.img or /boot/kernel7.img present
 if [ ! -f /rootfs/boot/${default_kernel_file} ] ; then
     echo -n "Copying kernel to /boot/${default_kernel_file}... "
-    cp /rootfs/boot/vmlinuz-${KERNEL_VERSION} /rootfs/boot/${default_kernel_file}
+    cp -- /rootfs/boot/vmlinuz-${KERNEL_VERSION} /rootfs/boot/${default_kernel_file}
     if [ $? -eq 0 ]; then
         echo "OK"
     else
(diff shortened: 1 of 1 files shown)


scripts/etc/init.d/rcS
```diff
@@ -787,7 +787,7 @@ do
 done
 
 # iterate through all the *.key files and add them to apt-key
-for keyfile in *.key
+for keyfile in $(ls *.key)
```
Mausy5043: `ls` shouldn't be used like that.

scripts/etc/init.d/rcS
```diff
@@ -764,10 +764,10 @@ do
 done
 
 # iterate through all the *.pref files and add them to /etc/apt/preferences.d
-for preffile in *.pref
+for preffile in $(ls *.pref)
```
Mausy5043: `ls` shouldn't be used like that. ref: http://mywiki.wooledge.org/ParsingLs

Mausy5043:
Just my €0.02

diederikdehaas:
Hi @hleitzell 
Thanks for finding some errors in the script :+1: 
I'm not clear however how most of your other changes are improvements.

In my initial implementation I did indeed use `$(ls *.pref)` but that was deliberately changed to NOT launch a subshell since there was no need and as @Mausy5043 pointed out there are actual problems with it. So why should we change it back?

The check on the return value of cp instead of checking whether the file exists was done to make it consistent with the rest of the code and afaik checking the existence of the file is not better. 
Is that a wrong assumption and if so, why?

Can you explain the benefit of doing `cp --` instead of `cp`?

diederikdehaas:
> Just my €0.02

Thanks @Mausy5043 :+1:

scripts/etc/init.d/rcS
```diff
@@ -739,12 +739,12 @@ old_dir=$(pwd)
 cd /rootfs/boot/config/apt/
 
 # iterate through all the *.list files and add them to /etc/apt/sources.list.d
-for listfile in *.list
+for listfile in $(ls *.list)
```
Mausy5043: http://mywiki.wooledge.org/BashPitfalls  
#1 on the list.

diederikdehaas:
> Can you explain the benefit of doing cp -- instead of cp?

Found the answer thanks to @Mausy5043's link: http://mywiki.wooledge.org/BashPitfalls#cp_.24file_.24target

scripts/etc/init.d/rcS
hleitzell: This is busybox POSIX sh and file globbing does not work if there are no files to match.  Doing this '$(ls *.key) in POSIX sh will produce no results instead of passing along '*.key' in keyfile.

scripts/etc/init.d/rcS
hleitzell: That is bash specific, not busybox sh.

hleitzell:
In the revert comments I wrote:
``
File globbing in busybox passes along '*' if there is no match and busybox does not create a subshell: http://git.busybox.net/busybox/tree/docs/nofork_noexec.txt
``
This is not bash, this is busybox POSIX sh which is not compatible.

hleitzell:
It could be rewritten to do a [ -f $var ] before any work but the 'ls *.{key,pref,apt}' works in POSIX as the link Mausy5043 points out at the top.

hleitzell:
As far as passing '--' to commands when doing filename expansion goes, it prevents accidental adding files that are valid in vFAT but could be parsed as arguments when pulled into Unix/Linux.

i.e. "apt/conf/internal-sources.apt" could be considered an '-s' option to 'cp' without the '--' indicating there are no more options to be parsed as detailed here: http://www.dwheeler.com/essays/fixing-unix-linux-filenames.html

diederikdehaas:
> It could be rewritten to do a [ -f $var ] before any work but the 'ls *.{key,pref,apt}' works in POSIX as the link Mausy5043 points out at the top.

That was indeed a fault in the old implementation and should be fixed.

I've written a short script to test things out (`/bin/sh` points to `/bin/dash` on my system):
```sh
#!/bin/sh
echo "Iterating of files without subshell"
for textfile in *.txt
do
    echo "   Processing textfile '$textfile'"
done
echo ""
for mp3file in *.mp3
do
    echo "   Processing mp3file '$mp3file'"
done
echo ""
echo "Iterating of files with subshell"
for textfile in $(ls *.txt)
do
    echo "   Processing textfile '$textfile'"
done
echo ""
for mp3file in $(ls *.mp3)
do
    echo "   Processing mp3file '$mp3file'"
done
```
The contents of the directory I tested it in:
```
$ ls -l
total 16
-rwxr-xr-x 1 diederik diederik 346 Feb 17 16:02 arraytest.sh
-rw-r--r-- 1 diederik diederik 107 Feb 18 01:13 config.txt
-rw-r--r-- 1 diederik diederik  95 Feb 17 15:59 domains.txt
-rwxr-xr-x 1 diederik diederik 427 Apr 29 19:07 filelistings.sh
```
Result:
```
$ ./filelistings.sh 
Iterating of files without subshell
   Processing textfile 'config.txt'
   Processing textfile 'domains.txt'

   Processing mp3file '*.mp3'

Iterating of files with subshell
   Processing textfile 'config.txt'
   Processing textfile 'domains.txt'

ls: cannot access *.mp3: No such file or directory
```
Assuming I have not made a mistake in my script, that means that with the subshell produces an actual error, while without the subshell it processes a non-existing file. The latter can easily be fixed with a check for existence of the file.

diederikdehaas:
> As far as passing '--' to commands when doing filename expansion

Yep, agreed. Hadn't thought of that. I had only seen the `--` construct before in relation to `start-stop-daemon`.
I think should probably quote the filenames as well to prevent issues with spaces.

hleitzell:
Yeah, I'll add onto this pull request with POSIX compliance across the board.  

I've been testing simply by installing busybox-static and doing a 'busybox sh' to validate the work.

Mausy5043:
Thanks! :+1: Great work.

diederikdehaas:
> Yeah, I'll add onto this pull request with POSIX compliance across the board. 

Thanks :+1: 

How about the revert for the `cp` statement?
You correctly identified a failed check for the result code, but I do prefer checking the result code over the existence of the file.

hleitzell:
I have a pull that does everything idempotently or fails instead of continuing (no check of return code required) which I prefer over an image that doesn't work.  I'll have to test it when I get home as I didn't have the time to do a full test last night.

diederikdehaas:
I usually make the distinction between fatal errors (the system just won't work) and non-fatal ones (the configuration isn't exactly as I intended), but I'll leave it up to you :-)

hleitzell:
https://github.com/debian-pi/raspbian-ua-netinst/pull/216 effecitvely does the same style of failing if something can not be done as I kept getting stuck with a Raspbian setup that I could not log into or sudo if I was missing a password.

diederikdehaas:
@hleitzell  I've cherry picked the commits related to this PR and fixed the merge conflicts.

PR #216, opened 2015-04-29
Correct user creation for remote SSH administration

hleitzell:
This allow to remotely connect to a setup via SSH for root and the custom user. It also modifies sudoers.d in case a password was not set so that sudo can still be accomplished.

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -590,7 +590,7 @@ if [ ! -z "$username" ]; then
         echo "OK"
         if [ -z "$user_passwd" ]; then
             echo -n "  Setting $username to sudo without a password... "
-            echo -n "%sudo         ALL = (ALL) NOPASSWD: ALL" > /rootfs/etc/sudoers.d/01_sudo || fail
+            echo -n "%$username ALL = (ALL) NOPASSWD: ALL" > "/rootfs/etc/sudoers.d/$username" || fail
             echo "OK"
         fi
     fi
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -533,66 +533,66 @@ fi
 echo "Starting install process..."
 cdebootstrap-static --arch=armhf $cdebootstrap_cmdline $release /rootfs $mirror --keyring=/usr/share/keyrings/raspbian-archive-keyring.gpg || fail
 
-# allow root login
 echo "Configuring installed system:"
-if [ ! "$disable_root" = "1" ]; then
-    echo -n "  Setting root password... "
-    echo -n root:$rootpw | chroot /rootfs /usr/sbin/chpasswd || fail
-    echo "OK"
-fi
 
-# add SSH key for root (if provided)
-if [ ! -z "$root_ssh_pubkey" ]; then
-    echo -n "  Setting root SSH key... "
-    mkdir -p /rootfs/root/.ssh && \
-        chmod 700 /rootfs/root/.ssh && \
-        echo "$root_ssh_pubkey" > /rootfs/root/.ssh/authorized_keys
-    if [ $? -eq 0 ]; then
+# configure root login
+if [ ! "$disable_root" = "1" ]; then
+    # add SSH key for root (if provided)
+    if [ ! -z "$root_ssh_pubkey" ]; then
+        echo -n "  Setting root SSH key... "
+        mkdir -p /rootfs/root/.ssh && \
+            echo "$root_ssh_pubkey" > /rootfs/root/.ssh/authorized_keys || fail
         echo "OK"
-    else
-        echo "FAILED !"
+        echo -n "  Setting permissions on root SSH directory... "
+        chmod -R go-rwx /rootfs/root/.ssh || fail
+        echo "OK"
+    fi
+    if [ ! -z "$rootpw" ]; then
+        # openssh-server in jessie doesn't allow root to login with a password
+        if [ "$release" = "jessie" ] ; then
+            echo -n "  Allowing root to login with password on jessie... "
+            sed -i 's/PermitRootLogin without-password/PermitRootLogin yes/' /rootfs/etc/ssh/sshd_config || fail
+            echo "OK"
+        fi
+        echo -n "  Setting root password... "
+        echo -n "root:$rootpw" | chroot /rootfs /usr/sbin/chpasswd || fail
+        echo "OK"
     fi
 fi
 
 # add user to system
 if [ ! -z "$username" ]; then
-    echo -n "  Creating user... "
-    chroot /rootfs /usr/sbin/adduser "$username" --gecos "" --disabled-password
-    if [ $? -eq 0 ]; then
+    echo -n "  Creating $username user... "
+    chroot /rootfs /usr/sbin/adduser "$username" --gecos "" --disabled-password || fail
+    echo "OK"
+    # add SSH key for user (if provided)
+    if [ ! -z "$user_ssh_pubkey" ]; then
+        echo -n "  Setting $username SSH key... "
+        ssh_dir="/rootfs/home/$username/.ssh"
+        mkdir -p "$ssh_dir" && \
+            echo "$user_ssh_pubkey" > "$ssh_dir/authorized_keys" || fail
         echo "OK"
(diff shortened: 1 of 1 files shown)


scripts/etc/init.d/rcS
```diff
@@ -533,66 +533,66 @@ fi
...
+        echo -n "$username:$userpw" | chroot /rootfs /usr/sbin/chpasswd || fail
+        echo "OK"
+    fi
+    if [ "$user_is_admin" = "1" ]; then
+        echo -n "  Adding $username to sudo group... "
+        chroot /rootfs /usr/sbin/usermod -aG sudo "$username" || fail
+        echo "OK"
+        if [ -z "$user_passwd" ]; then
+            echo -n "  Setting $username to sudo without a password... "
+            echo -n "%sudo         ALL = (ALL) NOPASSWD: ALL" > /rootfs/etc/sudoers.d/01_sudo || fail
```
diederikdehaas: I don't agree with this block.
Whether or not to ask for the password on a sudo permission is distinctly different from whether the user has a password set (which I personally would consider worthy of a `|| fail`, but that's another discussion).
Apart from that, I also don't want to encourage this **awful** use of sudo by having it hardcoded in the installer.
I realize that it's used (way too) often and if ppl really want to do that, they can provide it themselves through configuration files/scripts.

scripts/etc/init.d/rcS
```diff
@@ -533,66 +533,66 @@ fi
...
+# configure root login
+if [ ! "$disable_root" = "1" ]; then
+    # add SSH key for root (if provided)
+    if [ ! -z "$root_ssh_pubkey" ]; then
+        echo -n "  Setting root SSH key... "
+        mkdir -p /rootfs/root/.ssh && \
+            echo "$root_ssh_pubkey" > /rootfs/root/.ssh/authorized_keys || fail
+        echo "OK"
+        echo -n "  Setting permissions on root SSH directory... "
+        chmod -R go-rwx /rootfs/root/.ssh || fail
```
diederikdehaas: The `chmod 700` on the `~/.ssh` directory is better (and iirc that's what ssh checks for).
On various systems, I have files in the `~/.ssh/` directory which are world-readable and while your code doesn't seem to practically hinder that _now_, it does give the wrong impression regarding permissions of files in the `~/.ssh/` directory and could result in issues if code was added before your `chmod` statement.
I think the cleanest way for this code would be:
```sh
mkdir -p -m 700 /rootfs/root/.ssh || fail
echo "$root_ssh_pubkey" > /rootfs/root/.ssh/authorized_keys || fail
```
(and the same with the user creation)

diederikdehaas:
Nice improvements overall though :+1:

scripts/etc/init.d/rcS
hleitzell: I do not like the hardcode either but this is fairly standard for cloud computing such as with [Ubuntu cloud-init](https://cloudinit.readthedocs.org/en/latest/topics/examples.html) which produces:
```bash
vagrant@vagrant-ubuntu-vivid-64:~$ sudo cat /etc/sudoers.d/vagrant
# CLOUD_IMG: This file was created/modified by the Cloud Image build process
vagrant ALL=(ALL) NOPASSWD:ALL
```
Without a user password set and root disabled, the account can not be used to administrate the machine.

diederikdehaas: I realize that Ubuntu is (mostly) responsible for this awful behavior, doesn't make it less awful though :wink: 

> Without a user password set ... (and `ALL=(ALL) NOPASSWD:ALL`)

That's probably the biggest security hole I have ever seen, why not just `chmod -R ugo+rwx /` then? It's essentially the same.

scripts/etc/init.d/rcS
hleitzell: Those permissions are too open according to [OpenSSH's Guidelines](http://www.openssh.com/faq.html#3.14)

diederikdehaas: Then add 
```
chmod 600 /rootfs/root/.ssh/authorized_keys
```
(also recommended in _SSH Mastery_)

diederikdehaas: `ssh-copy-id` actually does what I stated, only with different code. 
It only applies `umask 077` to the creation of the directory and the `authorized_keys` file, but not recursively on all files. 
For example, you really want to have `id_rsa.pub` world-readable.

hleitzell: Got it.  I'll make it specific.

scripts/etc/init.d/rcS
hleitzell: Well, it's adopted by RedHat as well as Debian Jessie (hence the default root account disable).

diederikdehaas: I've just created a fresh VM with jessie and during installation it asks (and suggest you do) for a root password. 
If you do leave it blank, sudo is indeed setup, but **NOT** with the `NOPASSWD` switch.
And not using `root` and setting up the user without a password is NOT allowed.
```
diederik@deb-jessie-vm:~$ sudo cat etc/sudoers
#
# This file MUST be edited with the 'visudo' command as root.
#                         
# Please consider adding local content in /etc/sudoers.d/ instead of
# directly modifying this file.
#                         
# See the man page for details on how to write a sudoers file.
#                         
Defaults        env_reset
Defaults        mail_badpass
Defaults        secure_path="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"

# Host alias specification

# User alias specification

# Cmnd alias specification

# User privilege specification
root    ALL=(ALL:ALL) ALL

# Allow members of group sudo to execute any command
%sudo   ALL=(ALL:ALL) ALL

# See sudoers(5) for more information on "#include" directives:

#includedir /etc/sudoers.d
```

goranche: I personally like it this way,

    %sudo   ALL=(ALL:ALL) ALL
which allows you to just add the user to the sudo group...

and afterwards I disable root logins altogether... :wink:

hleitzell: So here is my problem.  I need to distribute the bootable unattended remote network installs to people who have a tendency to lose things.  Hard coding a password is a security issue since it can be logged in with via console if the SDcard is lost; hence the PKI using only SSH keys is preferable.

Creating a service account that can only be remotely logged into without a password required to sudo allows for hooking the unattended installer into a CI/CD setup to automate the rest of the installation and configuration.

Following the somewhat standard cloud methods to enable a service account for automation with sudo without compromising security is the best solution I could come up with.

diederikdehaas: Ok, that use case is a lot more nuanced then I thought. 
I thought that a user could login locally without any password and then get all the same privileges as root.
But the code _as is_ does allow just that and I'm not willing to accept that.

But AFAIUI you have actually given the proper example yourself:
```
vagrant ALL=(ALL) NOPASSWD:ALL
```
Which gives a **specific user**, namely `vagrant`, all rights without having to give a password ... and it's stored as a snippet in `/etc/sudoers.d/`. 
That is imo quite different from giving everyone in the sudo group all rights without a password and specifically with a **blank** user password.

What would be a good solution (I think) is:
- create a user with an impossible to guess password
- add the user's public key to `authorized_keys` so they can login (remotely) with PKI
- put a snippet in `/etc/sudoers.d/` which gives that specific user (or specific group) the rights it needs to operate (from for example `post-install.txt`)

Is that something that could work for your use case?

Now that v1.0.7 of the installer is released, I hope to put some (more) thought into this whole user creation stuff.
What I want is to extend it so that multiple users and groups can be created and users can (also) be placed in your own custom groups.

hleitzell: Looks like I didn't push up the latest changes I have at home which limit the sudo to the created user's group (ala cloud-init defaults).  I'll do that when I get home today hopefully so it updates the pull request.  

Since the adduser is passed "--disabled-password" the created user account is unable to be logged into via console at all unless the password is set later in the code (only happens if you have a password to set).

Currently I am using debian-installer / cloud-init to manage the initial device configuration for other hardware; I'm still interested in attempting to fix d-i for Raspbian when I have time as it takes care of all of this in a standardized way.

PR #217, opened 2015-05-03
Replace busybox-static with busybox

diederikdehaas:
This PR should **NOT** be merged before version 1.0.7 of the installer is released!

Primary reason for this PR is [the following comment on Debian bug 757941](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=757941#47):
_"More, I don't think reassigning it to glibc will do any good either,
because static linking has been discoraged there for ages."_

Here is a list of applets/programs which are not part of busybox, but are in busybox-static:
crond, crontab, dpkg, dpkg-deb, ed, fdisk, ifdown, ifup, passwd, su, sulogin, telnetd, tunctl

Of these I have replaced dpkg, dpkg-deb, fdisk, ifdown and ifup.
I think the other applets/programs are not needed in the installer itself and would likely make the installer a lot bigger in size.

This should finally fix issue #80 :-)


goranche:
> This should finally fix issue #80

oh come on, that one is becoming a tradition :grin: 

I'll will test this PR one day this week... quite interested in the result :+1:

Zeerix:
> Of these I have replaced dpkg, dpkg-deb, fdisk, ifdown and ifup.

What do you need `dpkg` for? It is not used in the installer except as `chroot /rootfs dpkg...`

diederikdehaas:
> What do you need dpkg for?

Even though the installer isn't (currently) using it directly, I can totally see that others use it in their `installer-config.txt`/`post-install.txt`/etc and we shouldn't break them.
Another reason is that it's listed as a dependency of various programs/packages.

I have left out others like `cron` because I don't see a use for them in the installer or the configuration files/scripts and if I thought wrong I expect them to file an issue for it.

diederikdehaas:
> This won't work

That's just a limitation of the busybox shell, right? 
Am I correct that that problem goes away when we switch to dash/bash as the shell?

Zeerix:
>> What do you need dpkg for?

> Even though the installer isn't (currently) using it directly, I can totally see that others use it in their  installer-config.txt/post-install.txt/etc and we shouldn't break them.
Another reason is that it's listed as a dependency of various programs/packages.

What would they use it for?
Well, I guess that unpacking *.deb files and then copying some parts over to /rootfs would be reasonable.

Zeerix:
> That's just a limitation of the busybox shell, right? 
> Am I correct that that problem goes away when we switch to dash/bash as the shell?

Yes. Or use a different busybox, like a custom build where this CONFIG_ is not set.

Mausy5043:
> I have left out others like cron because I don't see a use for them in the installer or the configuration files/scripts and if I thought wrong I expect them to file an issue for it.

Uhm. Hold on. Does that mean that this won't work anymore?

From my post-install.txt:

`chroot /rootfs /usr/bin/crontab -u $USER /$CRONTMP`

goranche:
> ```chroot /rootfs /usr/bin/crontab -u $USER /$CRONTMP```

this runs crontab in the installed filesystem (```/rootfs```), not from the installer (```/```)
from the installer, the path to the binary would be ```/rootfs/usr/bin/crontab```

Mausy5043:
:relieved:

goranche:
> thanks

no problem... chroot is a funny beast :wink:

Mausy5043:
yes very confusing if you don't pay attention.

diederikdehaas:
Busybox static from jessie is working now, thus this PR is no longer needed.

diederikdehaas:
@Zeerix I'm trying to rework this for v1.1.x  and I'd like to get your help/opinion on a few issues (others are free to join in too of course):

> >This will enable to install more full versions then the applets provided by busybox and which then take precedence over those applets.

> This won't work, because both busybox and busybox-static have CONFIG_FEATURE_PREFER_APPLETS set. Even if removing the symlinks in /opt/busybox/bin, the shell would not execute the replacement executable.

What happens to functionality which is present in busybox-static but not in busybox (like fdisk)?

> Am I correct that that problem goes away when we switch to dash/bash as the shell?

Does that also mean that a new PR without dash/bash and a changed interpreter in `scripts/etc/init.d/rcS` is therefor useless as it just won't work?

Zeerix:
> What happens to functionality which is present in busybox-static but not in busybox (like fdisk)?
If it's not a built-in applet, the shell will execute it from the file system.
> Does that also mean that a new PR without dash/bash and a changed interpreter in scripts/etc/init.d/rcS is therefor useless as it just won't work?
What do you mean?
Applets have preference, as long as busybox is your shell. If you use bash/dash as shell, you can use symlinks to busybox' applets selectively.

diederikdehaas:
> > Does that also mean that a new PR without dash/bash and a changed interpreter in scripts/etc/init.d/rcS is therefor useless as it just won't work?

> What do you mean?
The thought/assumption with that question was that the busybox shell would only find its own applets and fail if it wasn't there (no fdisk applet -> no fdisk functionality -> installer won't work since partitioning doesn't work).
But your other answer gave already the answer. In the case of fdisk it would then use the functionality of the fdisk package/program.
But I now understand that if there are 2 competing programs (ie with the same name), the applet always wins, regardless of the `$PATH` setting.

PR #218, opened 2015-05-04
Added notes about RPi2

josephernest:


diederikdehaas:

But as explained on IRC, 1.0.7 should be released shortly and that would make the issue go away.
Therefor I will not merge your PR.

Issue #176, opened 2015-01-08
Insecure validation of archive.raspberrypi.org signing key


mologie:
Hello,

the marked lines allows for malicious keys to be imported into the temporary keychain:
https://github.com/debian-pi/raspbian-ua-netinst/blob/master/update.sh#L115-L118

Using `gpg -k`, one merely checks for the presence of the good key in the previously imported file. It does not check for the absence of malicious keys. Specifically, `gpg --import` allows for importing multiple keys from one file. Therefore, an attacker must only append his malicious key to the file downloaded over HTTP (and leave the good key in it too) in order to inject a malicious key.

As fix, I recommend embedding a verified copy of the correct key into update.sh.

mologie:
This may probably count as duplicate of #64. I opened a new issue because the changes from #64 were already merged and the issue is very old (yet not closed).

diederikdehaas:
@mologie You make a valid and correct point and I specifically left issue #64 open since the raspberrypi.org key wasn't (and **still isn't**) verified. 
Consequently your recommended fix can not be applied. 
And it would also not [fetch the latest version of the key with `update.sh`](https://github.com/debian-pi/raspbian-ua-netinst/issues/64#issuecomment-46363867).
Plugwash has [provided proof](https://github.com/debian-pi/raspbian-ua-netinst/issues/64#issuecomment-46373413) that the raspbian.org key is the correct one though :smile: 

Apparently there is only 1 way to check a fingerprint before importing it, but that would require manually parsing gpg's output ... and I don't think that's a good way to go:
`gpg --with-fingerprint --with-colons FILE`
(source: http://lists.gnupg.org/pipermail/gnupg-users/2012-April/044078.html)

But ... I think I have found a proper solution by adding an extra check through counting the number of keys in the keyring: 
`gpg --homedir gnupg --with-colons --list-keys | grep "^pub:" | wc -l`
(source: http://lists.gnupg.org/pipermail/gnupg-users/2012-July/045044.html)
Thus after the first import, the count must be 1 and after the second it must be 2.

Note that I'm not very familiar with gpg stuff, so I can be wrong.

diederikdehaas:
Now that the raspberrypi.org signing key is downloaded over HTTPS, I think this issue is solved.
@mologie, what do you think?
See also commit https://github.com/debian-pi/raspbian-ua-netinst/commit/f4510092885d2c703309a45320347a4b153ad910

mologie:
[...]

PR #222, opened 2015-05-06
Verify the number of imported GPG keys

jturner314:
The `--import` command can be used to import keyrings in addition to individual keys, so it is necessary to check that only the two desired keys were imported. (Subsequent signature verifications depend on only valid keys being in the default keyring.)

The current version is vulnerable to attackers who can MITM HTTPS connections, because the attacker could replace `raspbian.public.key` with a keyring containing both `raspbian.public.key` and the attacker's key, then sign malicious files with their own key. The `update.sh` script would check that `raspbian.public.key` was imported but not notice the presence of the attacker's key. This commit prevents this vulnerability by ensuring that only the two desired keys were imported.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -106,31 +106,31 @@ download_package_lists() {
     curl -# -O https://archive.raspbian.org/raspbian.public.key
     gpg -q --homedir gnupg --import raspbian.public.key
     echo -n "Verifying raspbian.public.key... "
-    if gpg --homedir gnupg -k 0xA0DA38D0D76E8B5D638872819165938D90FDDD2E &> /dev/null ; then
-        echo "OK"
-    else
+    if ! gpg --homedir gnupg -k 0xA0DA38D0D76E8B5D638872819165938D90FDDD2E &> /dev/null ; then
         echo -e "ERROR\nBad GPG key fingerprint for raspbian.org!"
         cd ..
         exit 1
+    elif [ "$(gpg --homedir gnupg -k --with-colons | grep '^pub:' | wc -l)" -ne 1 ] ; then
+        echo -e "ERROR\nImported more than one GPG key for raspbian.org!"
+        cd ..
+        exit 1
+    else
+        echo "OK"
     fi
     echo -e "\nDownloading and importing raspberrypi.gpg.key..."
     curl -# -O https://www.raspberrypi.org/raspberrypi.gpg.key
     gpg -q --homedir gnupg --import raspberrypi.gpg.key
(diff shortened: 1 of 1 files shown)


diederikdehaas:
Hi Jim :+1: 

Does this mean my assumption in https://github.com/debian-pi/raspbian-ua-netinst/issues/176#issuecomment-99232585 was wrong?

What happens when a key is superseded? 
For example, currently the raspbian packages are signed with Mike Thompson's personal key and I know that @plugwash isn't happy about that. I don't know if/when it would be superseded with a 'Raspbian Archive Automatic Signing Key', but it would be rather unfortunate if then things would brake.

jturner314:
Hi @diederikdehaas :)

No, your assumption is correct as long as an attacker can't man-in-the-middle (MITM) the HTTPS connections used to download the keys (or temporarily alter the contents of the Raspbian/Raspberry Pi servers).

However, the `update.sh` script checks the fingerprints of the downloaded keys specifically to avoid having to trust the HTTPS connections. I just noticed that checking that the desired fingerprints are in the set of imported keys (which is all the script currently does) is insufficient to ensure that only the desired keys were imported.

If a key is superseded, its fingerprint in `update.sh` needs to be updated to the fingerprint of the replacement key. Until that happens, the `update.sh` script will terminate with the message:
```
ERROR
Bad GPG key fingerprint for raspbian.org!
```
or
```
ERROR
Bad GPG key fingerprint for raspberrypi.org!
```
This pull request doesn't change that, though. It would be possible to check if the replacement key was signed with the old key, but that seems like it would introduce unnecessary complication for right now.

By the way, this doesn't affect installed systems. They're using the `raspbian-archive-keyring` package, which should be updated with any new Raspbian keys. It only affects the `update.sh` script.

diederikdehaas:
Thanks for the explanation :+1:  
and you're of course right about the changed fingerprint on a superseded key :-)

Would it make sense to do a double check, first for a count of 1 after the raspbian.org key and then for a count of 2 after the raspberrypi.org key?

jturner314:
> Would it make sense to do a double check, first for a count of 1 after the raspbian.org key and then for a count of 2 after the raspberrypi.org key?

I don't think that's necessary, but it's a good idea and would give more feedback to the user. I just pushed an update that checks the count after each import.

diederikdehaas:
No, no need for that.

Thanks for your great contribution again :+1:

jturner314:
You're welcome. I'm happy to help.

goranche:
> I'm happy to help.

that's how good things get done :+1: :wink:

PR #223, opened 2015-05-06
Add PGP keys directly into repository

jturner314:
This simplifies the code substantially and makes it unnecessary to download the keys on each update. See issue #176.


diederikdehaas:
Since we'll keep downloading the GPG key, I'm closing this PR

PR #230, opened 2015-05-17
Fixed bug in copying wireless drivers

Zeerix:
On cpio build, the drivers for both kernels (rpi1, rpi2) were mixed up


diederikdehaas:
Seems better then what we had, thus merging.
Thanks :+1:

Issue #64, opened 2014-06-16
Fix insecure downloading of raspberrypi.org signing key


jturner314:
#62 provides verified installation for http://archive.raspbian.org packages, but the installer script also installs the http://archive.raspberrypi.org repository. In order to do so, it downloads the http://archive.raspberrypi.org signing key insecurely with `wget` and adds it to the trusted Apt keyring without any authenticity checks. This download is vulnerable to attack. The best solutions that I can think of are:
- Include a known-good copy of the key in the installer image so that it doesn't have to be downloaded during installation.
- Don't add the http://archive.raspberrypi.org repository during unattended installation, and let users know in the README that they should add it manually if desired and verify the key themselves. This repository doesn't have many packages outside of the http://archive.raspbian.org repository, anyway (list below).

```
8086tiny
8086tiny-dev
compoundpi-client
compoundpi-common
compoundpi-docs
compoundpi-server
icelib
icelib-dev
omxplayer
oracle-java7-jdk
oracle-java8-jdk
penguinspuzzle
pistore
pypy-setuptools
pypy-smbus-cffi
pypy-upstream
pypy-upstream-dev
pypy-upstream-doc
python3-lirc
python3-picamera
python3-pifacecad
python3-pifacecad-emulator
python3-pifacecommon
python3-pifacedigital-emulator
python3-pifacedigitalio
python3-pifacedigital-scratch-handler
python3-rpi.gpio
python3-snap-camera
python-lirc
python-picamera
python-picamera-docs
python-pifacecad
python-pifacecommon
python-pifacedigitalio
python-rpi.gpio
raspberrypi-artwork
raspberrypi-bootloader
raspi-config
raspi-copies-and-fills
rpi-update
simbamond
smartsim
sonic-pi
wolfram-engine
xserver-xorg-video-fbturbo
```

diederikdehaas:
The key used to sign the raspberrypi.org archive is now published securely on https://www.raspberrypi.org/raspberrypi.gpg.key
And ShitfPlusOne send me the following msg, just like plugwash has done earlier:

-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA1

I confirm the key being used to sign the raspberrypi.org repository is
pub   2048R/7FA3303E 2012-06-17
Key fingerprint = CF8A 1AF5 02A2 AA2D 763B  AE7E 82B1 2992 7FA3 303E
-----BEGIN PGP SIGNATURE-----
[...]

PR #233, opened 2015-05-21
Remove hardcoded wheezy in sources.list

bbinet:
This reverts commit 3e5323a3726f1b94a75499de139bd65562f2c8ea.

archive.raspberrypi.org now supports jessie.

Conflicts:
	scripts/etc/init.d/rcS

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -549,7 +549,7 @@ echo "OK"
 # add firmware to default apt sources
 echo -n "Configuring apt... "
 echo "deb http://mirrordirector.raspbian.org/raspbian $release main firmware" > /rootfs/etc/apt/sources.list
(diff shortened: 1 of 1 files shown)


diederikdehaas:
The directory structure has been made, but that's not the same as 'supports'.
If you look at the size and contents of http://archive.raspberrypi.org/debian/dists/jessie/Contents-armhf.gz you'll see it.

bbinet:
Oh you're right, I had just checked that apt-get update was working.

Btw, why do need to keep the archive.raspberrypi.org repository in the sources.list? all packages are not available in mirrordirector.raspbian.org?

diederikdehaas:
> all packages are not available in mirrordirector.raspbian.org?

Nope, see [here for a list](https://github.com/debian-pi/raspbian-ua-netinst/issues/64#issue-35763514)

bbinet:
Thanks for the information.

PR #234, opened 2015-05-21
Added gzip compression to installer.cpio files; improved build scripts

Zeerix:


goranche:
it would be better if you would split individual changes (makefile, gzip compression, other changes) into separate PRs, so they can be handled individually

Zeerix:
Ok, will do.

PR #235, opened 2015-05-21
Added Makefile

Zeerix:
Use 'make' to start ./update.sh and ./build.sh
Use 'make image' to start ./buildroot.sh
Use 'make clean' for ./clean.sh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,39 @@
+# Makefile for raspbian-ua-netinst
+
+# configuration
+BUILD_DEPEND = packages config scripts
+IMAGE_DEPEND = bootfs
+
+IMAGE = raspbian-ua-netinst-$(shell date +%Y%m%d)-git$(shell git rev-parse --short @{0}).img
+
+BUILD_DEPEND_FILES = $(shell find $(BUILD_DEPEND) 2>/dev/null)
+IMAGE_DEPEND_FILES = $(shell find $(IMAGE_DEPEND) 2>/dev/null)
+
+# build targets
+default: build
+
+update:
+packages: packages/updated
+build: bootfs/config.txt
+image: $(IMAGE)
+clean:
+
+.PHONY: default update build image clean
+
+# rules
+update:
+	@$(MAKE) --always-make packages
+
+packages/updated: update.sh
+	@rm -f packages/updated
+	./update.sh
+	@touch packages/updated
+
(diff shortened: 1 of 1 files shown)


goranche:
oh, you already did... sorry, didn't have time to comment on in #234 

anyway, I don't really see a benefit in using make for this, as it increases the number of required installed programs on the build machine...

I personally use a server to build by images, and as it is now, I don't need developer tools on it... ok, I could still manually invoke the scripts, but why have 2 methods?

Zeerix:
* "hey guys, how do I build it?" - "just type make"
* "make" is faster to write than ./update.sh && ./build.sh
* the make tool checks which files need an update; if you type "make image" it skips update.sh and build.sh unless you changed a file

Mausy5043:
Although I am hesitant to embrace the use of a `Makefile` (imho it overcomplicates things a bit) I do myself use a script to `git pull` etc. then `clean`, `update` and `build` a new installer. The script accepts one parameter (hostname) which is used for the `label=` (see #243) and the `hostname=` parameters set in the `installer-config.txt`.

Actually, I've completely abandoned modifying my own branch of raspbian-ua-netinst. Instead, I've migrated all modifications to `mod-raspbian-ua-netinst` (if you're interested look at: https://github.com/Mausy5043/mod-raspbian-ua-netinst). That way I don't have to keep in sync with raspbian-ua-netinst, I just grab the latest, overlay my own changes and start a build.

diederikdehaas:
I kind of like the idea to have a Makefile, but it should 'always' be an optional dependency.
Since I don't have much experience with Makefiles, it will take me a while to review the implementation (and I'm rather short on time atm).

goranche:
> I kind of like the idea to have a Makefile

why?

diederikdehaas:
Because it's often used on Linux systems to build a program/component/module.
On second thought maybe it's more because I (personally) want to learn more about Makefiles, which doesn't necessarily mean it's good for the project.
I do agree that having two ways to build the installer could be confusing and there's a real chance that at some points they (start to) diverge ...

goranche:
> it's often used on Linux systems to build a program/component/module

which we're not doing... at least not in the "make" sense of the term build, where you deal with building executables from source code...
my vote goes towards "no"

diederikdehaas:
Ok, I agree.
I still like to thank @Zeerix for providing me with a Makefile example :+1:

Issue #228, opened 2015-05-17
check for required utils


goranche:
The scripts don't check if required utils (zip for example) are available on the build machine.

Adding a code section (or separate script to be included) to do so should be quite straightforward :smile: 
In case a needed util doesn't exist, a message should be displayed to the user, ideally with instructions on how to install it.

diederikdehaas:
> should be quite straightforward

I'm not so sure about that, when you think of all the different systems this can be used on.
Patches welcome, but it's unlikely I will spend time on this.
What's wrong with `BUILD.md`?

goranche:
> when you think of all the different systems

I'm aware of this... that's why I added the "should be" :wink: 

I actually added this as a placeholder for some code I will be working on (and adding as a PR)...

diederikdehaas:
Ok :+1:

goranche:
huh... curl wasn't found on one of my servers :smile:

diederikdehaas:
Ping?

Mausy5043:
Closing. 
Feel free to re-open if required.

PR #236, opened 2015-05-21
Optimizied search for required packages in package list

Zeerix:
Reduces runtime of update.sh from 90s to 30s (on my test machine)
Requires bash 4.x

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -39,22 +39,23 @@ packages+=("libnl-genl-3-200")
 packages+=("libpcsclite1")
 
 required() {
-    if ((${packages_req["$1"]})); then
-        return 0
-    fi
+    for i in ${packages[@]}; do
+        [[ $i = $1 ]] && return 0
+    done
+    return 1
+}
+
+unset_required() {
+    for i in ${!packages[@]}; do
+        [[ ${packages[$i]} = $1 ]] && unset packages[$i] && return 0
+    done
     return 1
 }
 
 allfound() {
-    [[ ${#packages_req[@]} -eq 0 ]] && return 0
+    [[ ${#packages[@]} -eq 0 ]] && return 0
     return 1
 }
-
-# using an assotiative array is faster than looping through an array multiple times
-declare -A packages_req
-for i in "${packages[@]}"; do
-    packages_req["$i"]=1
-done
 
 filter_package_list() {
     grep -E 'Package:|Filename:|SHA256:|^$'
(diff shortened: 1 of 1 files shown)


goranche:
> Requires bash 4.x

not sure I'm happy with that... it's the reason things were done the way they are...
on the other hand, I don't really have a problem with saying "ok, you need bash 4"... although I would wan't some checks added in this case (would relate to #228 )

Zeerix:
Debian Squeeze and Wheezy already have bash 4. Not sure about Debian Lenny, but that is not in the repository anymore, is it?
But a check would be fine.

diederikdehaas:
I'm fine with requiring bash 4.
On [the GNU Bash homepage](http://tiswww.case.edu/php/chet/bash/bashtop.html#Availability) you can see that it's almost universally available on all platforms, except Solaris < 11 and MinGW but Cygwin does, so it should work on windows too.

(I've heard the installer also gets build on mainframes :stuck_out_tongue_closed_eyes: )

Mausy5043:
> (I've heard the installer also gets build on mainframes :stuck_out_tongue_closed_eyes: )
:wink: 
On the current version of OSX (10.10.3 Yosemite) people have:
```
$ bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin14)
```

On the 'mainframe' I have:
```
$ bash --version
GNU bash, version 4.3.30(1)-release (x86_64-pc-linux-gnu)
```

So no worries there, but for OSX users this might be a problem. Apple does not offer an upgrade to bash 4 (yet), so people will have to install a homebrew package called: `brew` to get to bash 4. Not sure if this is desirable.

Zeerix:
So I removed the associative array and executed `time ./update.sh`:

Original: `real    1m23.576s`
Patched: `real    0m32.515s`

diederikdehaas:
Nice :+1: 

(I'll review/test/merge it after the weekend unless @goranche beats me to it)

goranche:
> Anyone with OS X around?

yup :wink: 
(although I'm not building on OS X, too many problems for now, might have a go at changing the scripts some time in the future... but this is not a reason to stick with bash 3)

diederikdehaas:
@Zeerix Can you change the target branch to v1.0.x? 
You should be able to just modify this PR.
I/We will merge it into v1.1.x later on.

diederikdehaas:
Alternatively we can just merge this PR and create a backport for v1.0.x

diederikdehaas:
Thanks for your PR @Zeerix, much appreciated :+1: 

I'll do a backport and submit that as a separate PR

PR #237, opened 2015-05-21
Added copy&paste-able build requirements

Zeerix:


BUILD.md
```diff
@@ -13,6 +13,11 @@ On a Debian system those are the following, excluding packages with priority ess
 - dosfstools
 - binutils
 
+Install them as root:
+```
+aptitude install git curl bzip2 zip xz-utils gnupg kpartx dosfstools binutils
```
goranche: are we sure it will always be run on a debian compatible machine?

Zeerix:
Of course not, but this is just a convenience. The package names that are listed above can also be different on other distributions.

diederikdehaas:
`On Debian (based) systems you can install them as root or with sudo as follows:`
Seems more correct

diederikdehaas:
`bc` has been added recently to v1.0.x but we'll fix that when we merge v1.0.x into v1.1.x

PR #238, opened 2015-05-21
Adding gzip compression of installer cpio files

Zeerix:


Zeerix:
When gzipped, the files are about half the size. This leaves more space in /boot.

diederikdehaas:
I haven't tested it and I won't have time to test it the coming days, but doesn't this cause problems?
AFAIK the **kernel** itself knows how to handle cpio archives and that's quite important. 
That is not the case with gzip compressed files (afaik)

goranche:
I think the kernel will handle it if GZIP support was built into the kernel...

I think it is built in, but likewise won't have time to test it at least till monday... would be nice, though, as /boot will be around 8MB smaller

diederikdehaas:
I think I'll be ok with it:
```
$ file /boot/initrd.img-4.0.0-1-amd64 
/boot/initrd.img-4.0.0-1-amd64: gzip compressed data, last modified: Wed May 13 13:07:02 2015, from Unix
$ file dev/raspberry-pi/raspbian-ua-netinst/installer-rpi1.cpio 
dev/raspberry-pi/raspbian-ua-netinst/installer-rpi1.cpio: ASCII cpio archive (SVR4 with no CRC)
```
And from man:update-initramfs:
`The initramfs is a gzipped cpio archive.`

So yes, I would be fine with this.
(But I won't have time to test it till coming Tuesday)

goranche:
it would also probably make sense to put this into the v1.0.x tree...

Zeerix:
* I can back-port this (and the other changes) to the 1.0.x branch if you want; I just need the WLAN feature, so I develop on 1.1.x.
* the current default kernel used here can read cpio.gz, and it is common practice (i.e. the build scripts compress the initramfs image)
* why do we have two separate installer cpios? they are almost the same. the kernel modules lie in separate directories, so we could merge the initramfs files and save even more space.

goranche:
> we could merge the initramfs files and save even more space

I wouldn't go there at this time, as the image has problems between the different boards...

diederikdehaas:
I think it's a good idea to backport **this** PR to v1.0.x but not the others

Zeerix:
> I wouldn't go there at this time, as the image has problems between the different boards...

What do you mean? The installer initramfs are almost identical. Just the kernel modules are different.

diederikdehaas:

PR #239, opened 2015-05-21
Adding gzip compression of installer cpio files

Zeerix:
When gzipped, the files are about half the size. This leaves more space in /boot.

See also #238

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -203,9 +203,8 @@ function create_cpio {
     # zlib1g components
     cp tmp/lib/*/libz.so.1  rootfs/lib/
 
-    cd rootfs && find . | cpio -H newc -ov > ../installer-${target_system}.cpio
-    cd ..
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -205,6 +205,7 @@ function create_cpio {
 
     cd rootfs && find . | cpio -H newc -ov > ../installer-${target_system}.cpio
     cd ..
+    gzip -f --best installer-${target_system}.cpio
```
diederikdehaas: I don't like `-f` (force) unless it's absolutely necessary.

Zeerix: can be changed to:
    find . | cpio | gzip > file

diederikdehaas: Assuming you've used a shortened notation, then yes :+1:

diederikdehaas:

PR #243, opened 2015-05-28
Add a label to the bootpartition (v1.0.x)

Mausy5043:
Adding a label to the boot partition helps identifying the SD-card when plugged into a PC.


goranche:
would this be configurable from the config file? in that case it should probably be based off of v1.1.x :thought_balloon: 

(although, I'd also rather leave it in v1.0.x, and not document it here :angel:  what do you think, @diederikdehaas?)

Mausy5043:
Indeed, I use this for some time now. I modify the `label` in `installer-config.txt`.
I have a couple of SD-cards with different installs. The label tells me which is which. Easy-peasy really.
I was trying to make the same change for v1.1.x but something went wrong.

goranche:
> I was trying to make the same change for v1.1.x

there you can/should add it to the documentation as well :wink:

Mausy5043:
#245 is a better replica. :wink:

diederikdehaas:
I like the basic idea and am ok with having it undocumented in v1.0.x but documented in v1.1.x.
I do have a couple of remarks regarding the implementation though:
- `label` is a far too general word as a configuration parameter name, something like `boot_partition_label` is far more descriptive
- `raspberrypi` has that problem as well for the value. I usually name/label my partitions like `boot-part`, `root-part`, `var-part`, etc and I think that's far more descriptive. 
- There should be a check whether the configuration parameter actually has a value before using the `-n` parameter

There are 2 other things to keep in mind as well (but haven't made up my mind as to how to handle it):
- I like what [sjoerd has done with his image](http://sjoerd.luon.net/posts/2015/02/debian-jessie-on-rpi2/) by only placing the firmware files on the FAT partition (and mounting it as `/boot/firmware`) and put the rest of `boot` on an ext4 partition. I don't think the firmware packages support that (yet?), but I prefer not to make it impossible to support that when the packages do.
- There should also be an option to set the label on the `root` partition as far as the supported filesystems support it.

Mausy5043:
To label other partition one could use something like :
```
# e2label /dev/sda1 Boot
OR
# tune2fs -L Boot /dev/sda1
```
However, this was not the intent of this PR. The `label` is the volume name. It's primary use is to aid in easy recognition. Windows, OSX and Linux use the volume name of a VFAT partition as the text under the icon on the desktop. So, it helps in quickly identifying SD-cards by plugging them in an seeing what they are called.

Mausy5043:
> `raspberrypi` has that problem as well for the value

It's just a default. Override it in the installer-config.txt

Mausy5043:
> There should be a check whether the configuration parameter actually has a value before using the `-n` parameter

If the parameter has a default value there's no need to check. right?

goranche:
> OSX and Linux use the volume name of a VFAT partition as the text under the icon on the desktop

yup, IHMO this is the main thing you get with this

I'd personally like to have the build number (or short commit id) in the volume label :smile:

diederikdehaas:
> The label is the volume name. It's primary use is to aid in easy recognition.

I know and that's why I like the PR in principle. 
Apparently you can have different filesystems on a single partition, so maybe `boot-fs` or `pi-boot-fs` would've been a better suggestion. 
But I still think that `raspberrypi` is a poor default and I'd rather have an empty default value then.

> If the parameter has a default value there's no need to check. right?

Wrong. Maybe in some cases it's harmless, but in this case providing an empty label and not using the parameter have different effects (and maybe even causes an error, but haven't checked that).

goranche:
> If the parameter has a default value there's no need to check. right?

@Mausy5043 what if I, as a user, override it with an empty value in the config file? :wink:

Mausy5043:
@goranche @diederikdehaas : I've changed the default value and included a test for an empty string. Does that satisfy your issues?
We could discuss a different name for the parameter instead of `label` if you like.
How about `volumename` or `volumelabel`? As that is closer to what it is called in most of the documentation for VFAT-fs.

diederikdehaas:
It's a good improvement :+1: 

Regarding `label`, I primarily want to have `boot` in it to indicate it's the label/name/whatever for the `boot` partition/filesystem. That way we can apply similar things to other partitions/filesystems.

Mausy5043:
Can you agree with `boot-volumelabel` ?

diederikdehaas:
`boot_volume_label` sounds great to me :wink:

Mausy5043:
Anything else you'd like altered?

diederikdehaas:
I haven't had time to review it, but I hope to do it in the not too distant future.

diederikdehaas:
Thanks @Mausy5043 :+1:

PR #248, opened 2015-06-05
Fixed some tests in the installer.

Zeerix:
- Fixed a path
- Fixed test if copy was successful
- Removed unneccessary test for success

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -207,11 +207,7 @@ if [ "$drivers_to_load" != "" ] ; then
    do
       echo -n " Loading driver '$driver' ... "
       modprobe $driver || fail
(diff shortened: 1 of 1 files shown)


scripts/etc/init.d/rcS
```diff
@@ -207,11 +207,7 @@ if [ "$drivers_to_load" != "" ] ; then
    do
       echo -n " Loading driver '$driver' ... "
       modprobe $driver || fail
-      if [ $? -eq 0 ] ; then
-         echo "OK"
-      else
-         echo "FAILED !"
-      fi
+      echo "OK"
```
goranche: what if modprobe fails? is there a reason for removing this check?

diederikdehaas:
Thanks :+1:

PR #249, opened 2015-06-05
Improved support for rootfstype option

Zeerix:
* rootfstype can now be changed in online_config. previously, the mount options would be wrong for the new filesystem.
* if mkfs failed for the rootfs, the SD/microSD card was unbootable, because the bootloader was only copied back after that step. with this PR, the bootloader files are copied back earlier.


diederikdehaas:
Thanks :+1: 

I'll merge a backport to v1.0.x shortly.

PR #254, opened 2015-06-17
Add dependency to BUILD.md

goranche:
we use bc in buildroot.sh, but it's not installed on a Ubuntu server 15.04 system


goranche:
ugh... no... wrong merge target... :unamused:

PR #257, opened 2015-06-18
Add dash as shell interpreter

diederikdehaas:
This PR is dependent on https://github.com/debian-pi/raspbian-ua-netinst/pull/217 !

When https://github.com/debian-pi/raspbian-ua-netinst/pull/217 is merged, only the last 4 commits (starting with '[Added sensible-utils components.](https://github.com/debian-pi/raspbian-ua-netinst/commit/38a0c407f4bdc483462329eb1b2583906b9187be)') should be shown as commits specific to this PR.

[`dash`](http://manpages.debian.org/cgi-bin/man.cgi?query=dash&apropos=0&sektion=0&manpath=Debian+8+jessie&format=html&locale=en) is the standard command interpreter on Debian (based) system, by default symlinked to `/bin/sh`.
This means that the vast majority of scripts on a Debian (based) can be used as an inspiration for custom script run by the installer.
This also paves the way to add `bash` as command interpreter, since `dash` is a dependency of `bash`.


Zeerix:
Shouldn't a change like this be introduced in a new minor version, i.e. 1.1.x?

Some people might have a `installer-config.txt` that depend on the shell being a busybox.

diederikdehaas:
You do have a point.

OTOH the README.md says the following:
`The configuration file is read in as a shell script so you can abuse that fact if you so want to. See scripts/etc/init.d/rcS for more details what kind of environment your script will be run in (currently 'busybox sh').`
And the latter part, **currently 'busybox sh'** was intentional.

I'll have to think about this :thought_balloon:

diederikdehaas:
As it is right now, it shouldn't make a difference though.
Even though `dash` would be the default shell (`/bin/sh`), the installation script (`scripts/etc/init.d/rcS`) is still run with `busybox sh` and thus so would `installer-config.txt`.

But I did/do plan to change that at some point to `dash` (or `bash`), but that is not part of this PR and would be a separate one.

diederikdehaas:
Cancelled PR #217 thus this one is not applicable anymore.

PR #264, opened 2015-06-27
Move settings for /tmp to /etc/fstab so it works on all init systems.

diederikdehaas:


Mausy5043:
Surprised to see you apparently only need `nodev` and `nosuid`.
:-1:

scripts/etc/init.d/rcS
```diff
@@ -496,11 +496,8 @@ elif [ "$rootfstype" = "btrfs" ]; then
 else
 	echo "$rootpartition / $rootfstype $rootfs_mount_options 0 1" >> /rootfs/etc/fstab || fail
 fi
-echo "OK"
-
-# use ram tmpfs by default
-echo -n "  Configuring tmpfs to use RAM... "
-echo "RAMTMP=yes" >> /rootfs/etc/default/tmpfs 
+# also specify /tmp on tmpfs in /etc/fstab so it works across init systems
+echo "tmpfs /tmp tmpfs nodev,nosuid 0 0" >> /rootfs/etc/fstab || fail
```
goranche: adding noexec might make sense as well

goranche:
> Also no size-restriction?

I would leave that to the user

goranche:
> Surprised to see you apparently only need nodev and nosuid.

@Mausy5043 what surprises you about that? mount has defaults for most options anyway, so if you're ok with those, you don't need to list them specifically :wink:

diederikdehaas:
> Also no size-restriction?

I agree with @goranche to leave it up to the user. 
While it normally sets the size of tmpfs to 50% of RAM, it's not actually a problem, since [it's actual size is dynamically allocated according to the amount of data on /tmp](http://wiki.centos.org/TipsAndTricks/TmpOnTmpfs#head-9234b63540528fc2162ae32bf5a6939bbe43ae5d).

> adding noexec might make sense as well

I thought about that, but decided against it. See also the link above

PR #265, opened 2015-06-29
Merge latest updates from v1.0.x

diederikdehaas:


goranche:
many more, there were 
true-1

diederikdehaas:
Yeah, it looks messier than I expected.

PR #268, opened 2015-07-03
Added example for how to run script in target system context

botto:


README.md
```diff
@@ -89,7 +89,8 @@ So don't copy and paste the defaults from below!
 The _installer&#8209;config.txt_ is read in at the beginning of the installation process, shortly followed by the file pointed to with `online_config`, if specified.
 There is also another configuration file you can provide, _post&#8209;install.txt_, and you place that in the same directory as _installer&#8209;config.txt_.
 The _post&#8209;install.txt_ is executed at the very end of the installation process and you can use it to tweak and finalize your automatic installation.  
-The configuration files are read in as  shell scripts, so you can abuse that fact if you so want to. 
+The configuration files are read in as  shell scripts, so you can abuse that fact if you so want to.  
+If you want to write to the system that is being installed you will find it mounted under `/rootfs`
```
goranche: you might as well add that you can run commands in the installed system by doing ```chroot /rootfs cmd params```
and you also have ```/bootfs``` available...

:thought_balloon: on the other hand, I'd still refer anyone, who want's to take advantage of this, to look at [scripts/etc/init.d/rcS](https://github.com/debian-pi/raspbian-ua-netinst/blob/master/scripts/etc/init.d/rcS)

goranche:
thanks for your contribution :wink:

botto:
So I tried chroot and that stopped it from restarting, I didn't exit after chroot which I might have to do

goranche:
> I didn't exit after chroot which I might have to do

yes, you might :smile: 
if you don't, the installer will never be able to continue from the point of the chroot on

diederikdehaas:
I have never done an 'exit' after a chroot operations, so it surely isn't a requirement.
Can you post your statement, so that we can take a look?

goranche:
> I have never done an 'exit' after a chroot operations

not so much "exit" after a chroot operation (doing that would cause different issues), but the command that's run inside the chroot, needs to exit, before the installer script can continue with it's thing

diederikdehaas:
> but the command that's run inside the chroot, needs to exit

Precisely, hence my request to see the command

botto:
So I was trying to run

    chroot /rootfs /bin/bash
    cd /root
    wget https://github.com/gsliepen/tinc/archive/release-1.1pre11.tar.gz
    tar xf tinc-release-1.1pre11.tar.gz
    cd /root/tinc-release-1.1pre11
    ./configure
    make
    make install

This got stuck after the install was complete and asked me to hit enter to get into a terminal.

goranche:
uhm... this sequence of commands would try to install the tinc-release package onto the install image :wink: 

you'll have to write an install script onto ```/rootfs``` and run it there using chroot (instead of ```/bin/bash``` )

botto:
No...
When I ran this it does show up on the rootfs, the chroot has taken care of that

goranche:
you had that exact text in the ```post-install.txt``` file?

botto:
Your right, it writes to the build system, not sure why I wrote what I wrote like that.
Two things I've learnt:
1. Don't write a reply right before going to bed and very tired
2. Don't write scripts from memory, but look at what you wrote

What I had been using

post-install.txt:

    cp /bootfs/post-system-run /rootfs/root/post-system-run
    chroot /rootfs /bin/bash /root/post-system-run

post-system-run:

    cd /root
    wget https://github.com/gsliepen/tinc/archive/release-1.1pre11.tar.gz
    tar xf tinc-release-1.1pre11.tar.gz
    cd /root/tinc-release-1.1pre11
    ./configure
    make
    make install

Not sure why but earlier it was crapping out after the make install and asking me to go into terminal.
I ended up wiping the whole thing and starting from scratch again and it seems to work now, maybe I had some stray characters at the end of my post-install.txt causing it to die.

I'll update my PR in a minute with chroot instructions

diederikdehaas:
> This got stuck after the install was complete and asked me to hit enter to get into a terminal.
It did not indicate any errors, but simply asked me to go into a terminal

I'm a bit confused about this.
When things fail _during_ the installation, you often end up in the `fail()` function, which would then offer you to start a (rescue) shell ... for which you indeed need to press _\<Enter\>_ to enter it.

goranche:
> Your right, it writes to the build system

whew  :wink:

diederikdehaas:
HI @botto !

I do like what you've done and I think it would be a great addition to the _wiki_ but not the project (code) itself. So while I appreciate your contribution, I won't merge it and therefor will close this PR.
My idea is to move (pretty much) all the documentation over to the wiki and your example would be a great addition. :+1:

Issue #279, opened 2015-09-11
Base install using jessie fails when allowing root to login using password


mike632t:
Last entries in raspbian-ua-netinst-<timestamp>.log
```
Configuring installed system:
  Setting root password... OK
  Configuring /etc/fstab... OK
  Configuring tmpfs to use RAM... OK
  Configuring hostname... OK
  Configuring network settings... OK
  Enabling serial console... OK
  Allowing root to login with password... 
```
Fail routine catches error (didn't see message on screen untill just too late) and then system reboots and install process starts again.

'installer-config.txt' contains 
```
release=jessie
preset=base
hostname=pi-serv01
rootpw=*********
bootsize=+128M
rootsize=+1918M
cmdline="dwc_otg.lpm_enable=0 console=ttyAMA0,9600 kgdboc=ttyAMA0,9600 elevator=deadline"
```

diederikdehaas:
Sounds like https://github.com/debian-pi/raspbian-ua-netinst/issues/253

mike632t:
Thanks.

I have the installer running with serial console this time so won't miss anything (deleted ```console=tty1``` in cmdline.txt) will let you know if it works this time.

Couldn't find anywhere to just leave a coment - but this is an awsome installer!

diederikdehaas:
> Couldn't find anywhere to just leave a coment - but this is an awsome installer!

Thanks :+1:

mike632t:
Got a bit more from the console output this time - following should help.

```
  Allowing root to login with password... sed: /rootfs/etc/ssh/sshd_config: No such file or directory

Oh noes, something went wrong!
You have 10 seconds to hit ENTER to get a shell...
swapoff: /etc/fstab: No such file or directory
The system is going down NOW!
```

diederikdehaas:
Yeah, it's essentially the same as #253. `preset=base` and likely `preset=minimal` are broken (because I 'never' test them).

mike632t:
No issues with server (as you expected) - assume that to use the build instrustions (https://github.com/debian-pi/raspbian-ua-netinst/blob/master/BUILD.md)  you need to use a Pi ?

diederikdehaas:
> assume that to use the build instrustions (https://github.com/debian-pi/raspbian-ua-netinst/blob/master/BUILD.md) you need to run the build script on a Pi ?

No. I have never done that. I build the installer on my home PC running Debian Sid on amd64.

Issue #277, opened 2015-09-10
Hangs when rebooting after a 'base' install


mike632t:
The default installation options work fine - however in order to really minimize the disk space and allow me to use a serial console used I created an 'installer-config.txt'  containing the following.  

```preset=base
hostname=pi-serv01
rootpw=*********
bootsize=+128M
rootsize=+1918M
cmdline="dwc_otg.lpm_enable=0 console=ttyAMA0,9600 kgdboc=ttyAMA0,9600 console=ttyAMA0 elevator=deadline"
```

The installation appears to work (on a normal Pi with HDMI display) but the system does not come back up when rebooted, there seems to be something missing from the boot partition.  After the installation has completed if I delete all the files in the boot partition apart from  'installer-config.txt' and 'cmdline.txt'.  Then copy in all the files from the boot partition created using the default options the PI them boots normally.

diederikdehaas:
```
cmdline="dwc_otg.lpm_enable=0 console=ttyAMA0,9600 kgdboc=ttyAMA0,9600 console=ttyAMA0 elevator=deadline"
```
It looks to me that that line is wrong. I'm pretty sure that your console part should be `console=ttyAMA0,115200` and your 2nd `console` is wrong (AFAIK a serial console should have a speed) and you should just delete it. .
If you want both a 'normal' console (`tty1`) and a serial console which shows all boot messages, do something like this:
`cmdline="dwc_otg.lpm_enable=0 console=tty1 console=ttyAMA0,115200 kgdboc=ttyAMA0,115200`

Do you have a `/var/log/raspbian-ua-netinst.log` file? If so, could you post it?

mike632t:
I think the bit you want is this 

```
Installing bootloader package (=firmware)... OK
Installing libraspberrypi-bin package (=vcgencmd)... OK
Configuring bootloader to start the installed system...
Kernel version: ''
Copying kernel to /boot/kernel.img... cp: can't stat '/rootfs/boot/vmlinuz-': No such file or directory
FAILED !
Creating default cmdline.txt... OK
Saving current time for fake-hwclock...
Installation finished at Thu Sep 10 16:46:35 UTC 2015 and took 8 min 18 sec (498 seconds)
Unmounting filesystems... 
```

I can post the rest if you want

diederikdehaas:
That is indeed the part I needed. Something went wrong with the installation of the kernel and then the system can indeed not boot. Will look into that.
[...]

Issue #271, opened 2015-07-23
Pi2 stuck at 600mhz


ric96:
By default the pi2's clockspeed it stuck at 600mhz event if it is under 100% load.
Solution: install cpufrequtils

diederikdehaas:
Confirmed. 
The default scaling governor is _powersave_ and by installing `cpufrequtils` that changes to _ondemand_

kpfleming:
Thanks for reporting this! Giving up 33% of the CPU speed is not desirable :-)

5shekel:
thanks for this project, installed without a hitch. 
but would love seeing this installed (or mentioned) by default

PR #281, opened 2015-09-22
Fix presets

### LINKED ISSUE #253: Base preset bricks kernel installation
Hi,

when you choose the base preset to install a Raspberry PI 2 you'll always get a RSOD (Rainbow Screen Of Death) because it'll not install the module ${kernel_meta_package}.
In line 304 of /scripts/etc/init.d/rcS the variable cdebootstrap_cmdline will be completly overwritten.

Using the server preset and everything will be fine.

Greetings
Saij

diederikdehaas:
This PR re-implements the presets configuration so that a kernel is always installed as well as `cpufrequtils` to fully utilize the CPU.

This fixes #253, #277, #279 and #271.

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -305,7 +305,7 @@ if [ "$cdebootstrap_cmdline" = "" ]; then
     # not very logical that minimal > base, but that's how it was historically defined
 
     # base
-    base_packages="${kernel_meta_package}"
+    base_packages="${kernel_meta_package},cpufrequtils"
 
     # minimal
     minimal_packages="kmod,fake-hwclock,ifupdown,net-tools,ntp,openssh-server"
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -305,10 +305,10 @@ if [ "$cdebootstrap_cmdline" = "" ]; then
     # not very logical that minimal > base, but that's how it was historically defined
 
     # base
-    base_packages="${kernel_meta_package},cpufrequtils"
+    base_packages="${kernel_meta_package},cpufrequtils,kmod"
 
     # minimal
-    minimal_packages="kmod,fake-hwclock,ifupdown,net-tools,ntp,openssh-server"
+    minimal_packages="fake-hwclock,ifupdown,net-tools,ntp,openssh-server"
 
     # server
     server_packages="vim-tiny,iputils-ping,wget,ca-certificates,rsyslog,cron,dialog,locales,less,man-db"
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -301,40 +301,43 @@ esac
 # configure different kinds of presets
 if [ "$cdebootstrap_cmdline" = "" ]; then
 
+    # from small to large: base, minimal, server
+    # not very logical that minimal > base, but that's how it was historically defined
+
+    # base
+    base_packages="${kernel_meta_package}"
+
     # minimal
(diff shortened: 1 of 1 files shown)


scripts/etc/init.d/rcS
```diff
@@ -301,40 +301,43 @@ esac
...
+    base_packages="${kernel_meta_package},cpufrequtils,kmod"
 
-    # add latest kernel if rootfstype requires a module. 
-    # Changed it so it is always installed.
-    # Firmware can't be installed just yet, we'll do so at the end
-    if [ "$kernel_module" = true ] ; then
-        cdebootstrap_cmdline="$cdebootstrap_cmdline,${kernel_meta_package}"
-    fi
+    # minimal
+    minimal_packages="fake-hwclock,ifupdown,net-tools,ntp,openssh-server"
```
goranche: I'd add ```${base_packages}``` into the list here, not in the include parameter

goranche: although I'm not against it like it is, not to be misunderstood :innocent:

scripts/etc/init.d/rcS
```diff
@@ -301,40 +301,43 @@ esac
...
+        minimal)
+            cdebootstrap_cmdline="--flavour=minimal --include=${base_packages},${minimal_packages}"
             ;;
         *)
-            cdebootstrap_cmdline="$cdebootstrap_cmdline,vim-tiny,iputils-ping,wget,ca-certificates,rsyslog,cron,dialog,locales,less,man-db"
-            preset=server
+            # this should be 'server', but using '*' for backward-compatibility
+            cdebootstrap_cmdline="--flavour=minimal --include=${base_packages},${minimal_packages},${server_packages}"
+            if [ "$preset" != "server" ]; then
+                echo "Unknown preset specified: $preset"
```
goranche: maybe add "using server instead" to this echo?

scripts/etc/init.d/rcS
diederikdehaas: I had it like that at first, but changed it into what it is now. Don't recall right now what the reasoning of it was ...

scripts/etc/init.d/rcS
```diff
@@ -301,40 +301,43 @@ esac
...
+            fi
             ;;
     esac
 
-    # add extra packages
+    dhcp_client_package="isc-dhcp-client"
+    # add dhcp client if using dhcp
+    if [ "$ip_addr" = "dhcp" ]; then
```
goranche: should we play it safe and assume dhcp if ```${id_addr}``` is empty or even not set? (ok, not set isn't going to happen, I admit)

scripts/etc/init.d/rcS
goranche: cool

scripts/etc/init.d/rcS
diederikdehaas: We had it like this all the time. The only thing I did was re-organize the code a bit (and turned the hardcoded package name into a variable).

goranche: I know, but if it's changing, why not improve a bit as well :wink: 

it was just a suggestion

scripts/etc/init.d/rcS
diederikdehaas: Done (with slightly different words).

goranche: :+1:

scripts/etc/init.d/rcS
diederikdehaas: A bit late, but you're actually right :stuck_out_tongue_closed_eyes:

PR #289, opened 2015-10-08
Switch release to jessie

diederikdehaas:
Now that (even) the Raspberry Pi Foundation has switched to jessie, so should we (by default).


scripts/etc/init.d/rcS
```diff
@@ -304,8 +304,16 @@ if [ "$cdebootstrap_cmdline" = "" ]; then
     # from small to large: base, minimal, server
     # not very logical that minimal > base, but that's how it was historically defined
 
+    init_system=""
+    if [ "$release" = "jessie" ] ; then
+        init_system="systemd"
```
goranche: :disappointed_relieved:

PR #294, opened 2015-10-11
Fixed check on sources.list since 'listfile' starts with './'.

diederikdehaas:


scripts/etc/init.d/rcS
```diff
@@ -760,7 +760,7 @@ cd /rootfs/boot/config/apt/
 # iterate through all the *.list files and add them to /etc/apt/sources.list.d
 for listfile in ./*.list
```
goranche: I think it would be better to change this line to

```for listfile in *.list```

and leave 763 as is

diederikdehaas: https://github.com/debian-pi/raspbian-ua-netinst/pull/215#issuecomment-97498279 which points to http://www.dwheeler.com/essays/filenames-in-shell.html#globbing which explains why you should use `./*.list`

goranche: interesting... never noticed that  :thought_balloon:

goranche: of course, mainly because I always have additional checks, pretty much like in this case here

goranche: screen shot 2015-10-11 at 21 23 30

goranche: this is repeatable using both ```/bin/sh``` and ```/bin/bash``` (under Ubuntu 14.04.02 LTS)

to actually get the desired functionality (as described in the reasoning for using ```./*.list```), you'd have to use the following:

```for file in $(ls *.bla 2> /dev/null); do echo ${file}; done```

diederikdehaas: I don't know what point you're trying to make here.
I just noticed that a test which was expected to return false actually returned true and I fixed that.
If you don't like the way that the file iterations are implemented now, code away and submit a PR with your changes and an explanation why it is better.

goranche: uhm... :thought_balloon:

goranche: if you feel so strongly about it, go ahead and merge... I already commented what I think is wrong with it, you provided some counter argument which I've shown to be wrong and provided you with code that will actually do what you seem to be trying to accomplish, namely not run the loop if no file matches the glob...

PR #296, opened 2015-10-14
Replace busybox-static with busybox

diederikdehaas:
Re-introduction of a PR I submitted earlier (#217), but now rebased and targetted at v1.1.x

Primary reason for this PR is [the following comment on Debian bug 757941](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=757941#47):
_"More, I don't think reassigning it to glibc will do any good either,
because static linking has been discoraged there for ages."_

Here is a list of applets/programs which are not part of busybox, but are in busybox-static:
crond, crontab, dpkg, dpkg-deb, ed, fdisk, ifdown, ifup, passwd, su, sulogin, telnetd, tunctl

Of these I have replaced dpkg, dpkg-deb, fdisk, ifdown and ifup.
I think the other applets/programs are not needed in the installer itself and would likely make the installer a lot bigger in size.


goranche:
tested installation and all seems to work fine...

diederikdehaas:
:+1:

PR #301, opened 2015-10-25
Replace non-breaking hyphen ('‑' / &#8209;) by hyphen-minus ('-') for "installer-config.txt"

nagromc:
Copy and paste "installer‑config.txt" in the GitHub REAMDE.md page will
lead the installer to ignore the file since it expects
"installer-config.txt" instead.

Yeah, I'm a lazy person.


diederikdehaas:
Thanks :+1:

chris03:
Thank so much !
I couldn't understand why the installer would not use my custom settings.
I was copy/pasting the file name to be sure I didn't type it wrong but that was the problem !

PR #306, opened 2015-11-08
 build.sh: mkfs.btrfs path

EmilienM:
mkfs.btrfs binary is in */bin*, not in */sbin*.
Running build.sh was failing:
```bash
cp: cannot stat 'tmp/sbin/mkfs.btrfs': No such file or directory
```

With that change, it won't fail anymore.


EmilienM:
I did a mistake in my branch, I'll resubmit it.

PR #307, opened 2015-11-08
build.sh: fix mkfs.btrfs path

EmilienM:
mkfs.btrfs binary is in */bin*, not in */sbin*.
Running build.sh was failing:
```bash
cp: cannot stat 'tmp/sbin/mkfs.btrfs': No such file or directory
```

With that change, it won't fail anymore.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -145,7 +145,7 @@ function create_cpio {
 
 
     # btrfs-tools components
-    cp tmp/sbin/mkfs.btrfs rootfs/sbin/
+    cp tmp/bin/mkfs.btrfs rootfs/bin/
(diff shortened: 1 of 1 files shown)


diederikdehaas:
Then I'm guessing you have `release=stretch` in your `update.sh`, because on `jessie` the directory **is** `/sbin/`

EmilienM:
@diederikdehaas yes, I'm using a fork of your repo, with a custom branch, and I'm trying to build 'testing' images.
I guess you can close my PR and update the path in a future release :-)

diederikdehaas:
The workaround described here may work too: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=783672#22

A future version of the installer may switch to debootstrap as the jessie version of that program does support stretch.

Issue #166, opened 2014-12-08
older kernel version (3.2.0-4-rpi)


bbinet:
I was wondering if it was possible to build an image with an older raspbian kernel version (3.2.0-4-rpi).

The kernel version currently seems to be hardcoded in `build.sh`, `update.sh` and `scripts/etc/init.d/rcS`:

```
KERNEL_VERSION=3.12-1-rpi
```

Would it be possible to make the `KERNEL_VERSION` configurable?

Mausy5043:
The bug referred to seems to have been resolved.  
@bbinet : Is this still a issue?

diederikdehaas:
> The bug referred to seems to have been resolved.

That's not the impression I got. But I mostly considered this issue to be as a way for ppl to specify their own kernel package. Not sure that's wise, but it was the thought.

Mausy5043:
@bbinet : Is this still an issue with the newer kernel?

bbinet:
Sorry, I don't know, I'm now working with resinos for new boards.

Issue #194, opened 2015-03-08
Allow overriding the RPi version to install for


hadess:
After installing the system through QEmu:

```
$ cat /boot/config.txt 
[piUNKNOWN !!]
kernel=vmlinuz-
initramfs initrd.img- followkernel
```

It should be possible to override the detected Pi version through a command-line argument, so that the config.txt and image created is usable on a real Raspberry Pi.

```
raspbian-ua-netinst.hw=1
```

for example. Let me know if that's acceptable, I'll create a patch.

goranche:
this might be a good idea :thought_balloon: 
although probably not through a command-line argument, but an entry in the config file...

diederikdehaas:
How about defaulting to Pi 1?

diederikdehaas:
With the merging of #309 I think this issue is resolved.

PR #309, opened 2015-11-15
Add a way to specify the target hardware to install for

mutability:
This allows configuring the Pi version to target from the installer config, and also allows targetting both Pi1 & Pi2 together. (I am using the installer to build an installed image that will then be run on many different machines, so it needs to support both).

Should cover #194 too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ The format of the _installer&#8209;config.txt_ file and the current defaults:
     rootfs_mkfs_options=
     rootfs_install_mount_options='noatime,data=writeback,nobarrier,noinit_itable'
     rootfs_mount_options='errors=remount-ro,noatime'
+    hardware_versions=detect  # "detect" supports the install hardware only, set to "1 2" to produce an install that supports both Pi1 and Pi2
 
 The time server is only used during installation and is for _rdate_ which doesn't support the NTP protocol.  
 
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -35,7 +35,7 @@ online_config=
 usbroot=
 cmdline="dwc_otg.lpm_enable=0 console=ttyAMA0,115200 kgdboc=ttyAMA0,115200 console=tty1 elevator=deadline"
 rootfstype=ext4
-build_hardware_versions=detect
+hardware_versions=detect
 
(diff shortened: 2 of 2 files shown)


goranche:
interesting... need to test this :smile: 

and thanks for the contribution :+1:

diederikdehaas:
@mutability Could you add comments to the code blocks which explains what that code block is supposed to do? That would make it easier to follow what you're trying to do, what the logic behind it is, what the purpose is of each variable you're using, what (things) you're iterating over, etc.

mutability:
What needs clarifying?

mutability:
(I mean, I can add comments, but there's not much I can say that doesn't just reiterate what the code does)

diederikdehaas:
Maybe it's just too late now for me (UTC+1) ... let me get back to you tomorrow.

I am/was having problems with getting a clear picture of what was precisely happening at a certain point in the code. Now that I have done a successful install on a Pi 2 with "1 2" and a Pi 1 with "detect" :+1: it's probably (a lot) easier to determine exactly what happens where and why. I also had an urge to modify the code to add various "echo" statements in there, so I could better follow along.
Both are usually indicators that comments are useful, but again, I'm rather tired atm :wink: 

Another thing that may have 'clouded' my mind, was that I was myself thinking of renaming `kernel_meta_package` to `kernel_packages` to directly specify the kernel package(s) to install. That would also fix https://github.com/debian-pi/raspbian-ua-netinst/issues/166

goranche:
overall seems ok... there are some conflicts, though...

@mutability could you sync the pull request with the v1.1.x branch, so I can merge it?

diederikdehaas:
No need to, I'll fix the merge conflict (as I introduced it)

goranche:
> I introduced it

how could you?!? :scream: 

:wink:

Mausy5043:
Need to add an option for Pi0 too? :wink:

mutability:
I don't have a Zero but I think it uses the same broadcom SoC as the first-gen Pis so it'll probably just identify as a Pi 1. Is there anything that needs to be done differently on a Zero vs a Pi 1?

Mausy5043:
@mutability : guess you're right. I was kind of joking though. Given the fact that the Pi Zero has no ethernet, it'll be a challenge to get it far into the installation.

PR #310, opened 2015-11-15
Allow poweroff or halt after installing

mutability:
This lets you specify what to do at the end of the install (reboot / poweroff / halt).
Useful when using the installer to build an installed image; you don't want to actually boot into the installed system before you've taken a copy of the installed image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ The format of the _installer&#8209;config.txt_ file and the current defaults:
     rootfs_mkfs_options=
     rootfs_install_mount_options='noatime,data=writeback,nobarrier,noinit_itable'
     rootfs_mount_options='errors=remount-ro,noatime'
+    final_action=reboot  # set to poweroff to poweroff/halt at the end of the install rather than rebooting.
 
 The time server is only used during installation and is for _rdate_ which doesn't support the NTP protocol.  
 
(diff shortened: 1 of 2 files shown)


goranche:
another nice one, thanks :+1: 

will check out and test

goranche:
:thought_balloon: is there any real difference between poweroff and halt on the RPi?

mutability:
Poweroff blanks the console, halt doesn't.

goranche:
so no real difference :wink:

mutability:
No, there is a real difference.

It's useful to have the console stay on at the end of the install when you're debugging an install.

It's useful to have everything power down for unattended installs because you can run headless and just watch the Pi's lights.

diederikdehaas:
FWIW I think it's good to have both halt and poweroff as options.

Sorry for the delay @mutability, been really busy lately. But I'll probably get around to it this week.

diederikdehaas:
How difficult would it be to also apply this (possibly with an additional parameter) to `fail()` ?
When things now fail, you can end up in an endless failing loop because `fail()` always reboots.
Just a thought ...

diederikdehaas:
Works as advertised, nice work :+1:

Issue #42, opened 2014-04-14
Enable rng-tools by default


Asmod4n:
Since this is a headless install, entropy for ssh-host-keys will be very low, starting rngd before key creation should resolve that.

I just don't know if the default kernel has support for the raspi hardware rng.
I am rolling a self compiled kernel with hard-float support. (the kernel sources from https://github.com/raspberrypi/linux compile every arm kernel with soft-float)

Asmod4n:
This could be easily resolved by the kernel.. if it would block /dev/urandom on boot until it has enough entropy :/

kpfleming:
That's not how /dev/urandom works. Blocking there would break applications. If an application needs guaranteed amounts of entropy, it must use /dev/random.

Asmod4n:
Only Linux is showing this behavior, every other OS waits at boot for their non blocking random device to have enough entropy :/

kpfleming:
Even if that's true, this isn't the place to discuss it, since the team here has no responsibility for the Linux kernel :-)

diederikdehaas:
A VERY interesting thread on the raspberrypi.org forums: https://www.raspberrypi.org/forums/viewtopic.php?f=66&t=126892

Mausy5043:
How does one find out if the SSH keys were generated using the bcm2708-rng module or not?

diederikdehaas:
You can't (run a SSH key through a program to figure out which (if any) HWRNG was used).

For systems installed with this installer, none were generated with the assistance of bcm2708-rng unless the user explicitly (in `post-install.txt` for example) removed the generated keys, modprobe-d the module, installed rng-tools and started the service and _then_ ran `dpkg-reconfigure openssh-server`.

Mausy5043:
I've been doing some experiments and it seems that on the installed system `modprobe bcm2708-rng` is enough to enable the hardware RNG.

I need some kind of test to determine if the RNG in use actually **is** the hardware one or whether it really is just the software RNG that's active.

Also, is it necessary to install the `rng-tools` package? Or would it be enough to enable the kernel module?

Mausy5043:
Below some references I used to understand this subject. I'm parking them here for all to reference.

https://blog.cloudflare.com/ensuring-randomness-with-linuxs-random-number-generator/
https://www.howtoforge.com/helping-the-random-number-generator-to-gain-enough-entropy-with-rng-tools-debian-lenny
[...]

PR #311, opened 2015-11-15
Move the removal of cdebootstrap-helper-rc.d later, after post-install

mutability:
This helps if post-install wants to do package operations that would
usually invoke rc.d scripts.

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -735,11 +735,6 @@ echo ""
 echo "HWCLOCKACCESS=no" >> /rootfs/etc/default/hwclock || fail
 
 
-# remove cdebootstrap-helper-rc.d which prevents rc.d scripts from running
-echo -n "Removing cdebootstrap-helper-rc.d... " 
-chroot /rootfs /usr/bin/dpkg -r cdebootstrap-helper-rc.d &>/dev/null || fail
-echo "OK"
-
 # copy apt's sources.list to the target system
 echo "Configuring apt:"
 echo -n "  Copying raspbian repo to sources.list... "
@@ -885,6 +880,11 @@ if [ -e /bootfs/post-install.txt ]; then
     echo "=== Finished executing post-install.txt. ==="
     echo "================================================="
 fi
+
+# remove cdebootstrap-helper-rc.d which prevents rc.d scripts from running
+echo -n "Removing cdebootstrap-helper-rc.d... "
+chroot /rootfs /usr/bin/dpkg -r cdebootstrap-helper-rc.d &>/dev/null || fail
+echo "OK"
 
 # save current time if fake-hwclock
 echo "Saving current time for fake-hwclock..."


diederikdehaas:
The idea I have right now is actually to move that part 'upwards' in the code, precisely because of the ability to start services then. Most prominent example is starting the random number generator before installing the ssh-server (see https://github.com/debian-pi/raspbian-ua-netinst/issues/42).
Why do you think it would be a good idea to move it 'backward' ?

PS: Your other PRs look good at first sight :+1: , but have to test them and make sure they align with the direction we want to go. Haven't seen a problem yet, but do want to give it some thought.

mutability:
Why is it useful to start services while the system is half-installed?

The particular example I'm working with is installing piaware/dump1090 packages as part of the install phase; they feed ADS-B data from a USB dongle to a network server and it really doesn't make sense to start that as part of the install process. The dongle is probably not even there while installing, and if it is, it's likely that dump1090+piaware will start taking 90% CPU which is not great. So I really want to prevent the package install from running rc.d scripts.

If cdebootstrap-helper is removed before post-install runs then I will just have to reinstall it (or something equivalent) in post-install ..

mutability:
I guess the difference here is really that I am using raspbian-ua-netinst as a way to easily build installed images that will end up on different machines. The fewer things beyond the bare install that happen during install, the better for my usecase; things like seeding the RNG are not great in this case because I'm going to have to clean that up in postinstall to get it back to a pristine state suitable for booting on many machines (I already have to clean up a few things like ssh host keys)

diederikdehaas:
Ok, thanks for your insight. I still do want to enable the RNG stuff, but maybe we should enable/disable the cdebootstrap-help-rc.d before/after it. It won't matter in your case since you delete the ssh keys anyway, but for most ppl a good RNG would be useful, especially for generating the keys.

goranche:
> for most ppl a good RNG would be useful, especially for generating the keys

I agree with this... maybe add a setting "skip_all_services" (better name needed) that defaults to false, but may be set if needed?

diederikdehaas:
I'm torn about what to do with this PR.
In a way it makes sense to move it to the end and I think the current position is somewhat random.
OTOH it would solve _your_ problem, but it will introduce the opposite problem for people who 'depend' on being able to start services in `post-install.txt`. And I am one of those as I'm using `post-install.txt` precisely to experiment with starting the RNG service.

@goranche what do you think?

@mutability could you workaround the current behavior by running `chroot /rootfs /usr/bin/dpkg -i cdebootstrap-helper-rc.d` at the beginning of your `post-install.txt` ?

mutability:
Where do I get the package from during post-install? IIRC it's not a standard one, it's a magic one that cdebootstrap installs and which isn't in the main archives etc.

mutability:
You could just remove the package in post-install if you need to; that seems simpler.

diederikdehaas:
> Where do I get the package from during post-install?

It's part of `cdebootstrap-static`, so I _think_ you can just run the command I specified.

mutability:
dpkg -i wants a path; what is the path?

diederikdehaas:
`/usr/share/cdebootstrap-static/cdebootstrap-helper-rc.d.deb` 
You may have to copy that file over to the chroot first though

mutability:
Yeah, that's getting pretty horrible and fragile :(

diederikdehaas:
Just to be clear: I do think that a solution supporting both cases should be implemented.
I'm just not sure whether this (PR) is that solution or one we should adopt now.

diederikdehaas:
With "@goranche what do you think?" I actually had a different (type of) response in mind then just pressing the button... :-/

goranche:
hm... ok  :thought_balloon: 

I don't see a problem with moving the code to a later stage, if this can help someone... wan't me to revert it?

goranche:
but I do agree a better solution should be sought, but didn't think that should stop this PR  :blush:

PR #315, opened 2015-11-23
Security suggestion on swap permissions

dragetd:
Suggest removing the read-permissions on swap-file. Without, it was left world-readable, allowing other users to read the swap-file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ The latest kernel and firmware packages are now automatically installed during t
 When you need a kernel module that isn't loaded by default, you will still have to configure that manually.
 
 > Optional: `apt-get install raspi-copies-and-fills` for improved memory management performance.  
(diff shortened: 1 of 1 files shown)


diederikdehaas:
Do you have links as to why this would be needed?
I have never seen something like this, the [Debian wiki doesn't mention it](https://wiki.debian.org/Swap) and afaik the system takes care of its security.

dragetd:
If your /swap is world-readable, a local user can read the file and dump informations… strings, images whatnot. This can expose sensitive informations including key-material.

Simple test:
As a regular user:
> strings /swap
will output various interesting strings.

If this is not in the Debian wiki, it should be added there! :-P I'll talk to some Debian people…

goranche:
it's true that the swap file should be accessible only to root, but I'm not convinced that the installer creates the file this way...

goranche:
but I'll check, just to be sure

diederikdehaas:
It's only in the `README.md` :wink:

goranche:
ah... that would explain why I don't remember seeing anything resembling a swap file

diederikdehaas:
Reopened against `v1.0.x` (#316)

PR #324, opened 2015-12-07
Added check for existence of file before operating on it.

diederikdehaas:
The 'base' preset doesn't install the ssh-server package and thus also
doesn't have the sshd_config file.

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -565,7 +565,7 @@ echo "OK"
 
 # openssh-server in jessie doesn't allow root to login using password anymore
 # this hack does allow it (until a proper solution is implemented)
-if [ "$release" = "jessie" ] ; then
+if [ "$release" = "jessie" ] && [ -f /rootfs/etc/ssh/sshd_config ] ; then
     echo -n "  Allowing root to login with password... "
     sed -i 's/PermitRootLogin without-password/PermitRootLogin yes/' /rootfs/etc/ssh/sshd_config || fail
     echo "OK"


diederikdehaas:
@goranche The GitHub 'Merge pull request' button does an explicit `git merge --no-ff ...` while I always prefer to do a fast-forward merge. What do you think?
I haven't tried it, but I think I can also merge my branch with v1.1.x and if I do it (with ff) then both branches would contain the same commit (id). Otherwise you'd get various duplicate commit messages as you can see now in v1.1.x

goranche:
> Otherwise you'd get various duplicate commit messages

but you keep an intact history without any doubt on what happened when and where... it's pretty much the same reason I don't allow my minions to use rebase once a commit has been pushed to a non-local repository... FF will be done when it can be done anyway (I think, 95% sure)

goranche:
also, did this change? wasn't ssh-server in the base package once? :thought_balloon: I seem to remember something like that :blush: 

although, I have been using my custom post-install scripts for so long, I'm not even sure anymore

diederikdehaas:
> FF will be done when it can be done anyway (I think, 95% sure)

I am 100% sure that this PR can be done with a FF, but if you click the link `command line instructions` you see this:

**Step 1:** From your project repository, check out a new branch and test the changes.
```
git checkout -b diederikdehaas-check-existence-of-file-before-operating-on-it v1.0.x
git pull https://github.com/diederikdehaas/raspbian-ua-netinst.git check-existence-of-file-before-operating-on-it
```
**Step 2:** Merge the changes and update on GitHub.
```
git checkout v1.0.x
git merge --no-ff diederikdehaas-check-existence-of-file-before-operating-on-it
git push origin v1.0.x
```

diederikdehaas:
> also, did this change? wasn't ssh-server in the base package once?

No. You're mixing `base` and `minimal` (LOL)

`Other presets include minimal which has even less packages (no logging, no text editor, no cron) and base which doesn't even have networking.`

goranche:
> You're mixing base and minimal

meh, close enough... :smile: 

> this PR can be done with a FF

yes, it can be done, but I don't think it should be done with a FF, as this may lose history information (about the branch it was merged from)...

diederikdehaas:
> this may lose history information (about the branch it was merged from)

If/When this PR is merged, I'll delete that branch immediately, so I'm not sure anything is really lost. And not everyone uses a descriptive name for their branch like I did.
And the description of the PR is directly taken from the commit message and that does contain useful information, as it should. 
So I'm not convinced any (useful) information is lost.

goranche:
any good reason for using fast forward?

diederikdehaas:
It is the default and when I merge this branch also into v1.1.x they'll have the same commit id

diederikdehaas:
It could be that the 'mess' I'm seeing is caused by us/me merging v1.0.x into master and the other way around and then also merging into v1.1.x ...

goranche:
> It is the default

it obviously isn't (when using github, as well as other similar repo hosting solutions)...

diederikdehaas:
Thus if we **REQUIRE** that all PR are made onto v1.0.x (or v1.1.x) and NOT master, the problem may go away.

diederikdehaas:
> it obviously isn't (when using github, as well as other similar repo hosting solutions)

It's git's default. And I'm not 100% sure, but I think this is one of the reasons Linus thinks GitHub sucks for the git workflow (but it is good for purely hosting code).

goranche:
I still disagree... I like to keep the context of where particular commits came from intact...

diederikdehaas:
How about "we REQUIRE that all PR are made onto v1.0.x (or v1.1.x) and NOT master" (or manually merge them into v1.0.x or v1.1.x when someone still proposes it to master)?
Is that a good compromise?

goranche:
> all PR are made onto v1.0.x (or v1.1.x)

as far as I'm concerned that's already the case... and I'm not really sure what that would change in terms of merge type... but sure

diederikdehaas:
> that's already the case

It does say so in the `CONTRIBUTING.md` document, but not everyone reads that. 
And GitHub defaults to suggest things to merge into the default branch, which is master. 

> I'm not really sure what that would change in terms of merge type

It won't. But if we consistently merge v1.0.x into v1.1.x (and master on a new release), we may not get all those duplicates. I'm not sure, but I think it's worth a try.

PR #330, opened 2015-12-11
Fix hotplug device detection

diederikdehaas:


scripts/etc/init.d/rcS
```diff
@@ -102,10 +102,14 @@ export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/bu
...
-mount -t sysfs none /sys
+mount -t proc proc /proc
+mount -t sysfs sysfs /sys
 
-echo /sbin/mdev > /proc/sys/kernel/hotplug
+mount -t tmpfs -o size=64k,mode=0755 tmpfs /dev
+mkdir /dev/pts
+mount -t devpts devpts /dev/pts
+
+echo /opt/busybox/bin/mdev > /proc/sys/kernel/hotplug
```
goranche: is the ```/opt/busybox/bin``` right? :thought_balloon:

diederikdehaas: Yes. 
Because of `/bin/busybox --install /opt/busybox/bin/` ALL applets are in `/opt/busybox/bin/`.
Quite easy to check when the installer keeps crashing on you :stuck_out_tongue_closed_eyes:

goranche:
ok... didn't have the time to test it, so asking was faster :wink: 

will try it out, though, but probably tomorrow

scripts/etc/init.d/rcS
goranche: ok... didn't have the time to test it, so asking was faster :wink:

will try it out, though, but probably tomorrow

diederikdehaas: > will try it out, though, but probably tomorrow

:+1: 

I don't know if I've told you before, and otherwise it's now in the public domain, but I use this file for my `online_config` while having that file open in `vim` on my server:
```
echo "Hoi hoi from online-config.txt"
#wait_in_mins=10
#echo "Now we're going to wait for an arbitrary $wait_in_mins minutes to see whether the monitor stays on... "
#for i in `seq $wait_in_mins -1 1`; do
#	sleep 60
#	echo -n "$i.. "
#done
#echo ""
#echo "Let's continue with the installation!"
#echo "mounts:"
#mount
#which mdev
#echo "Let's crash the installer!"
#fail
```
This way I can easily test various things without having to rebuild the installer all the time.

PR #339, opened 2015-12-19
Allow installing entire gpg keyrings into /etc/apt

mutability:
This is useful if you want to manage the keyrings as separate files rather than bundling the keys into the main keyring.


diederikdehaas:
Nice :+1:

PR #340, opened 2015-12-19
Sleep before unmounting to make the logfile more likely to be complete.

mutability:
The problem is that the tee might not have finished writing out the end of the logfile when the copy/umount happens. I find it usually cuts off at least the "installation finished at .." line.

This is a bit of a bandaid - the "right" solution would be to preserve the original console fds at the start of the install, switch back to them at the end, close the pipe and wait for the tee to exit - but it's much simpler and seems to work OK.


diederikdehaas:
I've never seen that myself, but I'll take your word for it. 
OTOH I usually watch the installation via serial cable/console and then you wouldn't notice it.

diederikdehaas:
Thanks :+1:

PR #341, opened 2015-12-19
Add dosfstools to minimal packages for fsck.vfat.

mutability:
Without fsck.vfat, fsck of /boot is silently skipped on boot.


diederikdehaas:
Good addition (too) :+1: 
Any particular reason you've chosen to add it to `minimal` and not `base`?

mutability:
No particular reason, I just thought base was "stuff you absolutely have to have to get a bootable system". The system will boot OK without dosfstools, it just skips the fsck.

diederikdehaas:
Sounds like a proper reason to me :smile:

PR #342, opened 2015-12-20
Update README.md typo.

DavisMiculis:
Updated README.md SD card size typo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Other presets include _minimal_ which has even less packages (no logging, no tex
 
 ## Requirements
  - a Raspberry Pi Model 1B, Model 1B+ or Model 2B
- - SD card of at least 640MB or at least 128MB for USB root install (without customization)
+ - SD card of at least 64MB or at least 128MB for USB root install (without customization)
  - working Ethernet with Internet connectivity
 
 ## Writing the installer to the SD card


diederikdehaas:
Thanks for the idea :+1: 

But it's actually not a typo. The number came from a 'loose' addition of 128MB for `/boot/` and 512MB for `/` which is needed to install the system. The 64MB is only true for the installer image itself.

DavisMiculis:
Oh, sorry, just somehow while reading it felt like there should be 64 followed by 128. Maybe it just needs more clarification then?

diederikdehaas:
Yeah, that probably wouldn't hurt. I guess I came up with it because of Bill Gates' famous remark :wink:

Mausy5043:
@diederikdehaas Bill thought 640KB should be enough, not 64MB. :wink:

diederikdehaas:
You should at least correct me properly then ... not 640MB :wink: LOL

@Mausy5043 btw: interested in becoming an (official) Collaborator?

Mausy5043:
@diederikdehaas : Thanks for the offer. I need to think about this and have lots of questions. If you're still serious, could you contact me via LinkedIn maybe?

diederikdehaas:
Done :wink:

Issue #15, opened 2014-02-27
Can I specify a proxy?


GlowTube:
Trying to install jessie via ua-netinst... but I am behind a firewall and must specify an http proxy for external access.  Is there an option in installer-config.txt for this?

At the moment I get

E: Unknown suite jessie

but I suspect this is due to lack of external access, as folks have commented that jessie has been installed successfully via this method before.  Thanks!

bennettp123:
@diederikdehaas Worked fine for me. Thanks for your testing. :)

Mausy5043:
@diederikdehaas I've prepared PR #345 in the hopes of closing this issue.

Gurr3:
My code from installer-config
```
customproxy=proxyuser:proxypass@proxyserver:proxyport/
export http_proxy=http://$customproxy
export ftp_proxy=http://$customproxy
export https_proxy=https://$customproxy
```
which is then echoed into places in post-install
```
echo "http_proxy=http://$customproxy; export http_proxy" >> /rootfs/etc/bash.bashrc
echo "ftp_proxy=http://$customproxy; export ftp_proxy" >> /rootfs/etc/bash.bashrc
echo "https_proxy=https://$customproxy; export https_proxy" >> /rootfs/etc/bash.bashrc

echo "Acquire::http:Proxy \"http://"$customproxy"\";" >> /rootfs/etc/apt/apt.conf.d/50apt-proxy
echo "Acquire::https:Proxy \"https://"$customproxy"\";" >> /rootfs/etc/apt/apt.conf.d/50apt-proxy
echo "Acquire::ftp:Proxy \"http://"$customproxy"\";" >> /rootfs/etc/apt/apt.conf.d/50apt-proxy
```
its also included in rc.local to allow some automatic setup on first boot.

Note that using a password with special characters eg ( or space are required to be url encoded (%28, %20) with this solution

Mausy5043:
:+1: 

Have you tested this? Does the installer run from behind the proxy?
The reason I ask is that you seem to modify the *installed* system. So, I'm wondering how the installer would behave.

Gurr3:
Yes its tested, the installer runs with the changes made in installer-config, eg. `export http_proxy=...`.

The post-install parts is just to carry over into the installed system so it doesn't need to be reconfigured or updated later.

Mausy5043:
:cool: :+1:

diederikdehaas:
[...]

PR #345, opened 2015-12-25
Added short explanation for proxy-users

Mausy5043:
Seeking closure for #15

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,14 @@ Presets set the `cdebootstrap_cmdline` variable. For example, the current _serve
 
 (If you build your own installer, which most won't need to, and the configuration files exist in the same directory as this `README.md`, it will be include in the installer image automatically.)
 
+### Using a Proxy
+If you want to use a proxy, then simply add these lines to the `installer-config.txt`:
+
+    http_proxy=http://<proxy-server>:<proxy-port>/
+    export http_proxy
+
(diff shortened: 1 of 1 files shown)


diederikdehaas:
What's the reason for adding it to the documentation instead of adding a parameter and taking care of the implementation ourselves?

I am a bit reluctant to 'officially' add this to the project. I _think_ that proxy support can be rather tricky, because it can come in various 'flavors' and I don't have a (real) way of testing it.
The 'solution' as presented in #15 and this PR is therefor imo mostly a hack/workaround, which happen to work for that particular user. But proxies could require authentication and/or support for (some kind of) auto-configuration.

In a way I 'solved' the issue in a cowardly way by describing a possible workaround in the issue and keeping it open (closed issues are seldom found/searched). Also in the 'hope' that when another user has a need for proxy support, he/she would add their experiences to that issue. And then we would also have someone who could help troubleshoot proxy scenarios.

Wondering what to do and whether it may be more useful to copy the above text/info to issue #15 and continue the conversation there ...


2 (side) notes:
- The plan is to leave v1.0.x as is and (likely) only release a new version when a new kernel and/or firmware gets released and focus all development efforts onto the v1.1.x branch (@goranche and I had talked about this, but you couldn't know that)
- We always create topic branches in our own fork and then create a PR, just like everyone else. Not much of a problem now, but it's something to keep in mind for the future.

Mausy5043:
I'm OK with deleting this PR.

PR #346, opened 2015-12-27
Check for existence of locale files/programs before operating on them.

diederikdehaas:
Turns out it was only needed on files related to locales.
I thought also about `date +"%Y-%m-%d %H:%M:%S" > /rootfs/etc/fake-hwclock.data`, but decided against it as it won't generate an error (`fake-hwclock` may not be installed).

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -799,27 +799,35 @@ if [ "$locales" != "" ]; then
     echo -n "  Enabling locales... "
     for locale in $locales; do
         echo -n "$locale... "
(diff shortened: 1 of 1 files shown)


Mausy5043:
Reviewing this I just learned that `&>` is equivalent to `2&>1` which I always use. :+1:

diederikdehaas:
I actually think that `2&>1` is more descriptive then `&>` :thought_balloon:

Mausy5043:
`&>` is less bytes :stuck_out_tongue_winking_eye:

PR #348, opened 2015-12-28
Added 'drivers/net/usb' to support network devices over usb.

diederikdehaas:
There are various network devices which run over USB, besides wireless,
and they should now work too. Probably (also) useful for the Pi Zero.


Mausy5043:
Looks good to me.

diederikdehaas:
Just got word back from the person who triggered this PR and apparently the driver doesn't load automatically in the installer, but does on the installed system.
It probably makes sense to move the 2 driver lines **above** the `depmod` command.
It may not be it, but writing it down here now, so I won't forget the idea when I have time to test it.

PR #350, opened 2015-12-29
Fixed various shellcheck issues

diederikdehaas:
Implemented most suggesting that http://www.shellcheck.net/ indicated.
I left some items unresolved as they're rather debatable, which we of course can do :wink:

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -58,7 +58,7 @@ fail()
         mount $bootpartition /boot
         fail_boot_mounted=true
     fi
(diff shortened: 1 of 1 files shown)


diederikdehaas:
> pushd and popd still not in busybox?

I actually don't know. I had never heard of those commands when I wrote that part of the code.

Mausy5043:
I missed changes related to globbing and word splitting (https://github.com/koalaman/shellcheck/wiki/Sc2086)
But I guess that's one you want to debate? :wink:

diederikdehaas:
Yep, several of them are for counters. Others may have value, but they weren't as clear to me as the ones I already implemented. And you merged :+1:

Mausy5043:
:grin:

PR #352, opened 2016-01-03
Changed Pin condition to 'c=firmware' as that is a better differentiator.

diederikdehaas:
The 'n' (name) parameter doesn't distinguish between the raspbian.org repo 
and the raspberrypi.org repo as they both have wheezy and jessie.
But in the raspberrypi.org repo it is contained in the 'main' section
whereas it is in the firmware section of raspbian.org.
Bonus side effect is that if you switch to stretch (=current testing)
after having installed jessie first, then it still works.


Mausy5043:
It took me a while to understand what you're trying to do here. But I think I get it. Please don't ask me to explain though :grin:  

This helped me: http://linux.die.net/man/5/apt_preferences

diederikdehaas:
Do `apt-cache policy` and `apt-cache policy libraspberrypi-bin` and combine that with what you found in the manpage and you'll have another powerful tool in your toolbox :wink:

PR #360, opened 2016-02-13
HWRNG support for v1.1.x

### LINKED ISSUE #42: Enable rng-tools by default

Mausy5043:
Same as PR #359, now for the v1.1.x branch (This fixes #42)

[Review on Reviewable](https://reviewable.io/reviews/debian-pi/raspbian-ua-netinst/360)


diederikdehaas:
Ah damn, you had already finished this one, eh?

Mausy5043:
No, I want to give it one more test. I hope to do that this weekend.

diederikdehaas:
Ah ok :smile:

diederikdehaas:
Somewhat. But then you can still have the situation that the value is 1, but the `rng-tools` package still doesn't get installed.
That was probably the reason for my cumbersome description.
Not sure what the appropriate solution is or should be and/or whether we should do it as it seems to all based on assumptions.

Mausy5043:
OK, I'll look into it.

diederikdehaas:
> I prefer to default always to "enabled" and make a note in the documentation that those wishing to use a base system should consider whether they want to go with the default HWRNG or re-set it by themselves in the `installer-config.txt'

I think that is actually a good one :+1:

(doing too many things at the same time atm)

Mausy5043:
So, to summarize:

1. Added the package `rng-tools` to the installer
2. Added the module `bcm2708_rng` to the installer
3. Made the installer load the module (2) and start the daemon `rngd` from the package (1)
4. Added an option `hwrng_support=1` to the installer.
When `hwrng_support=1` : 
a. add the package `rng-tools` to the installed system via `base_packages`
b. let the installed system automagically load the module by adding it to `/rootfs/etc/modules`

scripts/etc/init.d/rcS
```diff
@@ -142,6 +143,15 @@ echo "================================================="
...
 echo "================================================="
 
+echo -n "Starting HWRNG "
+modprobe bcm2708-rng
+if [ $? -eq 0 ]; then
+    echo "succeeded!"
+else
+    echo "FAILED! (continuing to use the software RNG)"
+fi
+/usr/sbin/rngd -r /dev/hwrng
```
diederikdehaas: I would put this line directly after the `modprobe` so that the check for success also takes the starting of the daemon into account

Mausy5043: Hmm. How about making L147:
`modprobe bcm2708-rng && /usr/sbin/rngd -r /dev/hwrng`

diederikdehaas: I think having it on 2 separate lines is more clear

Mausy5043: but then the `if [[ $? ...` doesn't work.

Mausy5043: Then I should test `$?` each separately

Mausy5043: It's been changed

diederikdehaas:
Awesome work @Mausy5043 :+1:

PR #361, opened 2016-02-14
HWRNG support (v1.0.x) with `rng-tools` installed

Mausy5043:
Tests suggest that `rng-tools` is needed.

[Review on Reviewable](https://reviewable.io/reviews/debian-pi/raspbian-ua-netinst/361)


Mausy5043:
v1.0.x wont' be supported

PR #365, opened 2016-02-21
Provide dynamic linker in expected location

kpfleming:
Many executables copied into the initramfs are dynamically linked,
and have an embedded path to the linker/loader that should be used
to execute them. This patch adds a symbolicy link in that location
which points to the actual location in the initramfs.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -385,6 +385,10 @@ function create_cpio {
 
     # libc6 components
     cp tmp/lib/*/ld-*.so rootfs/lib/ld-linux-armhf.so.3
+    # some executables require the dynamic linker to be found
+    # at this path, so leave a symlink there
+    mkdir rootfs/lib/arm-linux-gnueabihf
+    ln -s /lib/ld-linux-armhf.so.3 rootfs/lib/arm-linux-gnueabihf/ld-linux.so.3
     cp tmp/lib/*/libanl-*.so rootfs/lib/libanl.so.1
     cp tmp/lib/*/libBrokenLocale-*.so rootfs/lib/libBrokenLocale.so.1
     cp tmp/lib/*/libc-*.so rootfs/lib/libc.so.6


diederikdehaas:
Do you have an example for this?
(Not that this PR would break anything, but it's good for me/others to be aware of it)

kpfleming:
Yep, if you add a ```/bin/sh``` line somewhere in rcS so you can get a shell during the installation process, try running ```ldd``` (or ```rdisc6```, added by one of the other PRs). The shell reports that they are 'not found', when they are clearly there. I used ```objdump``` to inspect the ELF headers of those programs after exhausting all other attempts, and noticed the hardcoded path. They work on the standard install because that's where the file is located, but it's in a different location in the initramfs.

diederikdehaas:
You are absolutely right :+1: 

Can you explain how you use `objdump`?
I am aware of [`objdump -p <program> | grep NEEDED`](https://lists.debian.org/debian-qt-kde/2014/09/msg00552.html) but I think I should know more about this command.
I just discovered that `/usr/sbin/rgnd` (the random number generator daemon) isn't working with the error message `not found` :wink:

diederikdehaas:
I prefer a slightly different implementation:
```patch
     mkdir -p rootfs/etc/network/{if-down.d,if-post-down.d,if-pre-up.d,if-up.d,interfaces.d}
+    mkdir -p rootfs/lib/arm-linux-gnueabihf/
     mkdir -p rootfs/lib/ifupdown/
```
```patch
     cp tmp/lib/*/ld-*.so rootfs/lib/ld-linux-armhf.so.3
+    cd rootfs/lib/arm-linux-gnueabihf
+    ln -s ../ld-linux-armhf.so.3 ld-linux.so.3
+    cd ../../..
     cp tmp/lib/*/libanl-*.so rootfs/lib/libanl.so.1
```
This way all the directories get created at the same place and symbolic links are created like the others (code consistency).

kpfleming:
```ldd``` (and ```pldd```) come from the ```libc-bin``` package, not ```binutils```. They are definitely present in the installer cpio archives.
Your script looks like it would create the proper link, but relative links can be fragile :-) If you execute ```/lib/arm-linux-gnueabihf/ld-linux.so.3``` from a shell prompt and it runs, then the link is good. As you can see from my patch, I prefer absolute links (and the link can also be made without switching directories, just include the target directory path in the link name as I did).

**Comment by diederikdehaas** [comment]
With the fix integrated, I still get:
```
ldd
/bin/sh: ldd: not found
which ldd
/usr/bin/ldd
/usr/bin/ldd
/bin/sh: /usr/bin/ldd: not found
ls -l /lib/ld-linux-armhf.so.3
-rwxr-xr-x    1 1000     1000        134448 Feb 21  2016 /lib/ld-linux-armhf.so.3
ls -l /lib/arm-linux-gnueabihf/ld-linux.so.3
lrwxrwxrwx    1 1000     1000            24 Feb 21  2016 /lib/arm-linux-gnueabihf/ld-linux.so.3 -> /lib/ld-linux-armhf.so.3
```
But it has solved another issue we were having. Thoughts?

kpfleming:
That's quite odd; I'll test it again with the fix merged and see what I can find. The unfortunate part is that ```ldd``` is the tool generally used when debugging this sort of problem...

diederikdehaas:
Feel free not to spend time on this as I don't think a working ldd in the installer is that important :wink:

Mausy5043:
@diederikdehaas But for debugging purposes it would be very convenient :wink:

diederikdehaas:
`ldd` and `objdump` are very convenient tools and I use them often, especially when adding new components to figure out if/which components need to be added as dependency. I just never use it in the installer itself, but on my main host PC and/or a `chroot`.

Issue #347, opened 2015-12-28
IPv6 support


kyle95wm:
Hi,

I just tried to install this release using a 2GB SD card. I downloaded the image and flashed it to my SD card and I seem to have encountered a boot loop of sorts...

The error Im getting could be seen here: http://prntscr.com/9jbwol

What is going on exactly? Why am I getting this error?

diederikdehaas:
Do you have a combination of IPv4 and IPv6 in your local network?

Mausy5043:
Or a VM?  192.168.101.0/24 is typically used on VLANs.
Please provide us with more context. What does your network look like? 
What equipment must the data pass to get from the "the cable in the street" to your Pi?

kyle95wm:
Yes my network currently uses both IP 4 and 6. The network ID mentioned in the other comment was correct. Should I be disabling ip 6?

diederikdehaas:
The reason I thought of that was the following message on IRC:
```
Long story… but indeed… messed up ipv6 config. Router was advertising ipv6, but the routing failed. And debian would not fallback to ipv4. Were unable to fix it, disabled everything, now it worked
```
So it looks like the 'quick fix' is indeed disabling IPv6.
I don't know whether the installer can deal with IPv6 properly or that it is your (and the other user's) router config is not functioning 100%.

kyle95wm:
Ive disabled IP6 on my network and everything seems to be working fine. Is there not a way to make the instal fall back to IP4 if IP6 fails?

diederikdehaas:
Maybe. But I have hardly any experience with IPv6, so it may take a while (my router does support IPv6, but my ISP doesn't).
Also, this issue _could_ have pointed out an issue with your IPv6 configuration. Not saying it does, just that it could as it was the case of that other user.

Mausy5043:
Why use IPv6 for your LAN anyway?

diederikdehaas:
Because you can? Or experimenting with it to learn about it?

dragetd:
Why use a total legacy protocol like IPv4? :-P

diederikdehaas:
LOL. We still have to support those people who are stuck with legacy ISPs, like me :stuck_out_tongue_closed_eyes:

Mausy5043:
@kyle95wm : is this still an issue for you? 

It is not entirely clear to me if the IPv6 support of this repo is "incomplete" or whether your router might be "defective".

goranche:
[...]

PR #366, opened 2016-02-21
Add 'resolvconf' package to 'minimal' preset

kpfleming:
With IPv6 support in place, there can be two dynamic sources of
recursive DNS server information in the system. To handle this,
install the 'resolvconf' package which will coordinate updates.

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -400,7 +400,7 @@ if [ "$cdebootstrap_cmdline" = "" ]; then
     fi
 
     # minimal
-    minimal_packages="fake-hwclock,ifupdown,net-tools,ntp,openssh-server,dosfstools"
+    minimal_packages="fake-hwclock,ifupdown,net-tools,ntp,openssh-server,dosfstools,resolvconf"
 
     # server
     server_packages="vim-tiny,iputils-ping,wget,ca-certificates,rsyslog,cron,dialog,locales,less,man-db"


diederikdehaas:
Why not (something like) [`rdnssd`](https://packages.debian.org/jessie/rdnssd) and is it possible to make it conditional on when IPv6 is actually used?

diederikdehaas:
I just made a post in #347 which is also relevant to this and I'd appreciate your input on that :smile:

kpfleming:
The full IPv6 patch will add ```rdnssd``` to the package list as well, but only when IPv6 is enabled (and automatic nameserver discovery is requested). We could make the patch only add ```resolvconf``` when IPv6 is enabled, but I was concerned about having no-IPv6 and yes-IPv6 installed systems behave differently and require different configuration after installation (if the user wants to change how nameservers are configured). Since resolvconf is fairly small, and IIRC all shell scripts, it seemed like a safe thing to add.

Without ```resolvconf```, we'd have to create our own method of coordinating between ```isc-dhcp-client``` (for IPv4) and ```rdnssd``` (for IPv6) in the final installed system.

diederikdehaas:
> resolvconf is fairly small, and IIRC all shell scripts

My worry was that it was a running service and would therefor also use RAM. Turns out this is not the case and that relieves me of my main concern.

diederikdehaas:
> The full IPv6 patch

I haven't yet merged your other PRs as I'd like to see it in full.
(and it actually would be easier if they're all in 1 PR)

kpfleming:
Fair enough, I'll get my branch updated against the current master branch, and then send a PR. I'll keep the individual commits so they can be commented on separately if desired.

diederikdehaas:
Excellent, I do like separate commits :+1:

PR #367, opened 2016-02-21
Ensure 'ip' commands run by udhcpc only affect IPv4 addresses

kpfleming:
The udhcpc script uses various 'ip' commands to manipulate the IP
addresses during installation; this patch ensures that the commands
only affect the IPv4 (family 'inet') addresses and routes, thus
avoiding removal of IPv6 addresses added by the kernel or other
processes.

### DIFF
--- a/scripts/etc/udhcpc/default.script
+++ b/scripts/etc/udhcpc/default.script
@@ -19,7 +19,7 @@ PEERDNS_IF=eth0
 case "$1" in
   deconfig)
     # bring interface up, but with no IP configured:
-    ip addr flush dev $interface
+    ip -f inet addr flush dev $interface
     ip link set $interface up
     # remove any stored config info for this $interface:
     rm -f $CFG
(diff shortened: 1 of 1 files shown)


kpfleming:
This one should be considered independently of the IPv6 PR, since it doesn't add any software to either the installer or the installed system. It's just fixing a bug in the script.

diederikdehaas:
Fair enough :wink: 

I'll test it out shortly.

scripts/etc/udhcpc/default.script
```diff
@@ -19,7 +19,7 @@ PEERDNS_IF=eth0
 case "$1" in
   deconfig)
     # bring interface up, but with no IP configured:
-    ip addr flush dev $interface
+    ip -f inet addr flush dev $interface
     ip link set $interface up
```
diederikdehaas: I should probably experiment with it myself (some more), but you may already know it by heart:
why does this line not have a `-f inet`?

Otherwise, it's looking very good and I will merge your PR today

goranche: because the link command affects the network device, so protocol families don't make sense here...

scripts/etc/udhcpc/default.script
```diff
@@ -19,7 +19,7 @@ PEERDNS_IF=eth0
 case "$1" in
   deconfig)
     # bring interface up, but with no IP configured:
-    ip addr flush dev $interface
+    ip -f inet addr flush dev $interface
```
goranche: I personally prefer the `-4` flag (which is a shortcut for `-family inet`... :thought_balloon:  it's more intuitive (for me, at least)

scripts/etc/udhcpc/default.script
diederikdehaas: Thanks, makes a lot of sense :+1: 
The man page wasn't exactly clear on this ... and as I still don't have IPv6 myself, I couldn't test. (I should fix that though). So there was/is the possibility that it would be possible for a dual-stack-ed device.

scripts/etc/udhcpc/default.script
diederikdehaas: Yep, I agree

goranche:
also, after merging, this should be added to the v1.0.x branch as well :thought_balloon:

diederikdehaas:
Excellent, thanks a lot @kpfleming :+1:

PR #368, opened 2016-02-21
Add 'rdisc6' to initramfs for IPv6 DNS server discovery

kpfleming:
IPv6 networks commonly use RDNSS, inside Router Advertisements,
to communicate the addresses of recursive DNS servers to hosts
on the network. This patch adds the rdisc6 tool so the installer
script can obtain the addresses.

A follow-up pull request will change the script to actually use the tool.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -297,6 +297,9 @@ function create_cpio {
     cd ../../../..
     cp tmp/usr/sbin/arpd rootfs/usr/sbin/
 
+    # ndisc6 components
+    cp tmp/bin/rdisc6 rootfs/bin
+
     # lsb-base components
     cp tmp/lib/lsb/init-functions rootfs/lib/lsb/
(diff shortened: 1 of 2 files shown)


diederikdehaas:
Looking at https://packages.debian.org/jessie/ndisc6 is looks like you also need to add `perl` to the installer.

kpfleming:
Thankfully, no :-) There are some Perl scripts included in the ndisc6 package, but we're not copying them into the installer. ```rdisc6``` is a plain binary, with no special dependencies. I was happy to find that out, because otherwise I'd have to find another solution... adding perl to the installer would have increased its size substantially.

diederikdehaas:
Ah ok, phew :+1:

kpfleming:
This will be included in the combined PR.

Issue #357, opened 2016-01-30
documentation should be updated


oupala:
The paragraph [Reinstalling or replacing an existing system](https://github.com/debian-pi/raspbian-ua-netinst/#reinstalling-or-replacing-an-existing-system) of the main README file seems not to be up to date.

I can't find `kernel_install.img` in my boot partition. I can only find `kernel-rpi1_install.img`, `kernel_install.img` and `kernel_install.img`.

Same problem for `installer.cpio.gz` which is not present in my boot partition. I can only find `installer-rpi1.cpio.gz` and `installer-rpi2.cpio.gz`.

The README file should be adapted to say that the filename should be adapted depending on which raspberry you are using.

diederikdehaas:
You're absolutely right, thanks for bringing it to our attention :+1:

diederikdehaas:
Fixed with the merge of https://github.com/debian-pi/raspbian-ua-netinst/pull/369

PR #369, opened 2016-02-21
Fixing documentation

Mausy5043:
* typos
* whitespace (automagically fixed by Atom)
* issue #357
* increased readability of the options descriptions

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ There are currently 2 ways in which you can contribute:
 This document outlines how to best report issues you may encounter and the way we prefer pull requests.
 
 ## Reporting issues
-First, search through the existing [issues](https://github.com/debian-pi/raspbian-ua-netinst/issues) to see whether your issue is already known. If it is, add your details to the existing report. The more details we have, the better the chance for a fix. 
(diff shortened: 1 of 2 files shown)


Mausy5043:
There are still some issues with the `README.MD` that don't feel right. It would be great if someone who is a native speaker could check for grammar and dutchisms and the like.

kpfleming:
I have some networking-related additions to make as part of the IPv6 work, so I'm happy to do a pass over the whole README.md once this is merged.

Mausy5043:
:+1: thanks

diederikdehaas:
Great @Mausy5043 and thanks in advance @kpfleming :+1: 

> whitespace (automagically fixed by Atom)

Be VERY careful with this! Whitespace in markdown is significant. 
For example 2 whitespaces translate into a break/linefeed.
A quick look at the changes you made here didn't show any formatting breaking, but automatic formatting tools can easily have effects you didn't intend.
At some point I want to transform the documentation from GitHub flavored markdown to pure markdown and then it becomes (even) more significant.

Mausy5043:
> For example 2 whitespaces translate into a break/linefeed.
This I did not know.
Thanks for the heads-up.

diederikdehaas:
I/we had an earlier PR which automatically 'fixed' various space issues ... but that one did have negative consequences :wink:

diederikdehaas:
Thanks @Mausy5043 :+1:

PR #370, opened 2016-02-21
Installer options sorted alphabetically

Mausy5043:
Increases readability


diederikdehaas:
While I agree that the ordering of the config options can/should be improved, I don't think that sorting them alphabetically is the solution.
It's a good idea to figure out how we like to group config options. One group that immediately springs to mind is network, but we should figure out appropriate groups for all of them.

Not closing this PR just yet as it is a useful discussion subject.

Mausy5043:
Agree.

Mausy5043:
I propose to close this PR.
If needed we can open an issue about ordering the options differently.

diederikdehaas:
Agreed :+1:

PR #371, opened 2016-02-21
Changed default bootoffset to 8192.

### LINKED ISSUE #283: Sandisk Ultra 64GB Micro SDXC (Class 10) doesnt work
I tested it with 2 cards, both doesnt work with your image. Working flawless with the latest Raspian Image. What might cause this trouble? Missing FW?

### LINKED ISSUE #299: SD partioning should be changed
The SD card is formatted from the first available block. This is a really bad idea on SD cards. The first erase block (usually 4M) has firmware stuff on it. If it dies, so does the whole card. It isn't wear leveled so is extra fragile. This is why almost all Raspbian images start their first partition at block 8192. I killed a lot of compact flash cards on x86 systems a few years ago before I knew this...

I tried adding variables to rcS to fix this, but while a test busybox script worked on my laptop to fdisk the sd, it failed on a ua build, it sat there and did nothing, no log, either.
I tried this:
bootoffset=8192 used for first boot partition block
rootoffset=$((`echo $bootsize | sed s/+// | sed s/M//` * 2048 + $offset )) used for first rootpartition block when bootdev=rootdev, otherwise it starts at bootoffset

I gave up and just set rootoffset=270336 after a few goes. In hindsight, could make partition 3 which ends at 8191 first, then do /boot and /root, then delete 3. Should work?

diederikdehaas:
This fixes #299 and possibly fixes #283.


diederikdehaas:
Is there anything missing in this PR?

Mausy5043:
No there isn't. Just got distracted by some random stuff ;-)

diederikdehaas:
Cheers :+1:

PR #372, opened 2016-02-21
Moved mkdir statement to the other mkdir statements.

diederikdehaas:


kpfleming:
+1

Mausy5043:
Are you sure this is a smart `mv`? ;-)
Taking code apart and grouping similar bits of code together doesn't feel right. Doesn't this make it more difficult to read the code later on?
Could you elaborate on your motivation for this PR?

diederikdehaas:
Can you explain why having mkdir statement scattered all over the place is in any way better?

Mausy5043:
Well, IMHO code that is set-up in a modular fashion is easier to read and maintain. I would `mkdir` the directory, immediately followed by any activity that requires the presence of that dir.

Disclaimer: my opinion in no way prescribes my own behaviour :wink:

goranche:
I agree with @Mausy5043 on this one...

diederikdehaas:
And if I hadn't grouped all those mkdir statements where they are now, you would've found them all over the place.
Example:
I want to add `zlib1g` components which places files in `rootfs/lib/` so should the first command in that section be `mkdir rootfs/lib/`? 
What to do if I then want to add the components for `libuuid1`? Or `libtinfo5` or `lsb-base` or ... ?

diederikdehaas:
Feel free to motivate your opinion so others can learn ...

Mausy5043:
> And if I hadn't grouped all those mkdir statements where they are now, you would've found them all over the place.

Correct. And I see no problem with that.

> Example: .....
Well, now you made me think and regret what I said before. :disappointed: : 

When I looked at the rest of script I could see that the proposed change conforms to the layout of the script and therefore that would not be a valid reason to hold the merge ... IMHO. Furthermore, agreeing with the proposed change in itself I merged.

In general, I like code to be modular. The title of the PR triggered me to ask you that question. 
But that's just my general rule of thumb and in no way carved in stone.
It's just that I think that code benefits (or rather people trying to read the code benefit) from code that is written in such a way that related lines are place in each others vicinity. 
I should not have used this PR to push my opinion and for that I apologise.

diederikdehaas:
> In general, I like code to be modular.
> It's just that I think that code benefits (or rather people trying to read the code benefit) from code that is written in such a way that related lines are place in each others vicinity.

Couldn't agree more. But right now, the code isn't modular **at all**.
In the absence of modularity, I strive for (some) structure and consistency _as the next best thing_.

IMO creating all the needed directories in one place is at least a bit more structured then having `mkdir` statements all over the file in random places.
That's also the reason why I suggested [in PR 365](https://github.com/debian-pi/raspbian-ua-netinst/pull/365#issuecomment-186728151) to change the creation of the symbolic link as that was more consistent then how it is implemented now. Not that it better, but after experiencing several faulty created symbolic links before, I had settled on the way I proposed it. 
Can it be improved? Undoubtedly . 
And when someone has a better approach to do them, we should then fix **all** of them.

> I should not have used this PR to push my opinion and for that I apologise.

Don't _ever_ apologize for giving your opinion. That's also one of the things I like about OpenSource and Debian: everything happens in the open. And by doing that, everyone can learn and jump in when they want to.

goranche:
> so should the first command in that section be mkdir rootfs/lib/

why not? and if you add the `-p` option, you can run it as often as you want (answering the other question, what to do if you want to install several stuff, that places things in the same dir)

diederikdehaas:
> why not? and if you add the -p option, you can run it as often as you want

Because you'd get a WHOLE lot of the **same** `mkdir` statements.

goranche:
so?

diederikdehaas:
Sounds rather redundant and error prone

goranche:
what do you think that .deb packages do? they all create directories as needed, using the `-p` flag, so no error is raised if the dir already exists (and parent dirs are created as needed)... what's error prone?

redundant? sure... but it makes much more sense when reading the code, and makes maintaining the code (refactoring) **much** easier

goranche:
still... I don't understand how you think this makes the code error prone? :thought_balloon:

diederikdehaas:
> what do you think that .deb packages do?

So you like to equate this installer with the APT system? Yeah, I'll take that compliment :grinning:

goranche:
> you like to equate this installer with the APT system

uhm... no? :smirk:

PR #379, opened 2016-03-05
Remove obsolete line

Mausy5043:
How did this one get by us! Daemon is already started in L148


diederikdehaas:
That might explain why I was getting the following error when testing another PR:
`can't lock /var/run/rngd.pid, running daemon's pid may be 105`

PR #380, opened 2016-03-05
Run alternative rcS, if present

goranche:
If a `rcS` file is found in the config directory of the SD card, the installer (`rcS`) script will execute that file instead of itself.
This allows for much easier installer development, as there is no need to rebuild the installer image every time a change needs to be tested.

As a side effect, it also allows end users to add to (or change) the installer script without a need to rebuild the image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,12 @@ Presets set the `cdebootstrap_cmdline` variable. For example, the current _serve
 
 (If you build your own installer, which most won't need to, and the configuration files exist in the same directory as this `README.md`, it will be include in the installer image automatically.)
 
+### Custom installer script
+
+It is possible to replace the installer script completely, without rebuilding the installer image. To do this, place a custom `rcS` file in the config directory of your SD card. The installer script will check this location and run this script instead of itself. Take great care when doing this, as it is intended to be used for development purposes.
+
+Should you still choose to go this route, please use the original [rcs](https://github.com/debian-pi/raspbian-ua-netinst/blob/master/scripts/etc/init.d/rcS) file as a starting point.
+
 ## Logging
 The output of the installation process is now also logged to file.  
 When the installation completes successfully, the logfile is moved to /var/log/raspbian-ua-netinst.log on the installed system.  
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -121,7 +121,7 @@ klogd -c 1
 # Check if there's an alternative rcS file and excute it
 # instead of this file. Only do this if this isn't the
 # alternative script already
-if [ -z ${am_subshell} ]; then
+if [ -z ${am_subscript} ]; then
     mkdir -p /boot
     mount $bootpartition /boot
     if [ -e /boot/config/rcS ]; then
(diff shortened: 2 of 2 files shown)


goranche:
I tested this by placing the original (including the changes in this PR) onto the SD card (in the `/config` directory), and the installer worked fine... there were 3 warning messages (well, individual error messages), but those were expected... these relate to mounting already mounted filesystems (`proc`, `mdev`) and creating a link that already exists (`/bin/sh`)

kpfleming:
This looks very useful.

goranche:
@diederikdehaas yes, we could use that path as well, I guess... fixing

also, it's easier if you comment on the changes in the PR, not on my individual commit, as the comment actually shows up at the line of code without the need to go into the commit ;)

diederikdehaas:
> the comment actually shows up at the line of code without the need to go into the commit

Not sure if I get you. Do you want me to put remarks in the 'Files changed' tab of the PR?

goranche:
> put remarks in the 'Files changed' tab of the PR?

yes, that sounds much clearer... :blush:

diederikdehaas:
I can do that :+1:

goranche:
path was fixed...

and if you'd commented on the PR itself, that comment would've collapsed :wink:

scripts/etc/init.d/rcS
```diff
@@ -118,6 +118,30 @@ mdev -s
...
 klogd -c 1
 
+# Check if there's an alternative rcS file and excute it
+# instead of this file. Only do this if this isn't the
+# alternative script already
+if [ -z ${am_subscript} ]; then
+    mkdir -p /boot
+    mount $bootpartition /boot
+    if [ -e /boot/config/installer/rcS ]; then
+        cp /boot/config/installer/rcS /rcS
```
diederikdehaas: Is it wise to store the file directly in the root of the filesystem (vs in a subdirectory of it)?

scripts/etc/init.d/rcS
```diff
@@ -118,6 +118,30 @@ mdev -s
...
+        echo "---------------------------------------------"
+        echo "=== Execution of alternative rcS finished ==="
+        echo "============================================="
+        ${final_action} || reboot || exit
+    else
+        # Clean up, so the rest of the script continues as expected
+        umount /boot
+    fi
+fi
+
```
diederikdehaas: Any particular reason you've put this before the logging is setup?

scripts/etc/init.d/rcS
goranche: it's as good a place as any :thought_balloon: 

have any particular reason in mind? it's a RAM based filesystem, usually you might even find an init script here...

scripts/etc/init.d/rcS
goranche: I'd put this even earlier, if possible... the point is to "replace" the original rcS and be able to do development work on the alternative file (so not to build the image every time a change is made), so the less is done by the original rcS file, the better...

also, the logging would be overridden by the alternative rcS file anyway

README.md
```diff
@@ -168,6 +168,12 @@ Presets set the `cdebootstrap_cmdline` variable. For example, the current _serve
 
 (If you build your own installer, which most won't need to, and the configuration files exist in the same directory as this `README.md`, it will be include in the installer image automatically.)
 
+### Custom installer script
+
+It is possible to replace the installer script completely, without rebuilding the installer image. To do this, place a custom `rcS` file in the config directory of your SD card. The installer script will check this location and run this script instead of itself. Take great care when doing this, as it is intended to be used for development purposes.
```
bbinet: It seems the `rcS` file should be located in the `installer` subdirectory, and not directly in the `config`, right?

PR #381, opened 2016-03-05
Add check for existing lo interface

goranche:
It seems Wheezy adds a lo interface to `/etc/network/interfaces`, this PR adds a check, which prevents the script from adding another one in such case

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -528,7 +528,13 @@ echo "OK"
 # default hostname
 echo -n "  Configuring hostname... "
 echo -n $hostname > /rootfs/etc/hostname || fail
-echo "127.0.0.1	localhost" > /rootfs/etc/hosts || fail
+if [ ! -f /rootfs/etc/hosts ]; then
+    echo "127.0.0.1	localhost" > /rootfs/etc/hosts || fail
+else
+    if ! grep -q "localhost" /rootfs/etc/hosts; then
+        echo "127.0.0.1	localhost" >> /rootfs/etc/hosts
+    fi
+fi
 if [ "$domainname" = "" ]; then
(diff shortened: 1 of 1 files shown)


scripts/etc/init.d/rcS
```diff
@@ -528,7 +528,13 @@ echo "OK"
 # default hostname
 echo -n "  Configuring hostname... "
 echo -n $hostname > /rootfs/etc/hostname || fail
-echo "127.0.0.1	localhost" > /rootfs/etc/hosts || fail
+if [ ! -f /rootfs/etc/hosts ]; then
+    echo "127.0.0.1	localhost" > /rootfs/etc/hosts || fail
+else
+    if ! grep -q "localhost" /rootfs/etc/hosts; then
+        echo "127.0.0.1	localhost" >> /rootfs/etc/hosts
+    fi
```
Mausy5043: I think you shouldn't test for "localhost" but for "127.0.0.1", because in Debian "127.0.1.1." also redirects to "localhost". :thought_balloon:

diederikdehaas: > in Debian "127.0.1.1." also redirects to "localhost"

No, it redirects to the hostname, which is not the same as localhost (although both on the same machine)

Mausy5043: oh yeah you're right.
`::1` redirects to localhost

goranche: IPv6 isn't included here yet, no... :wink: 

also, this is a very simple check to see if a package that was installed, already created the `localhost` entry; so a simple check for the presence of a `/etc/hosts` file and if `localhost` exists in it, should be more than enough...

goranche: even if `localhost` is set to something different than `127.0.0.1` :thought_balloon:

Mausy5043: :OK:

PR #382, opened 2016-03-06
elevator=noop

Mausy5043:
Use `elevator=noop` i.s.o. `elevator=deadline` because `noop` is more efficient for SD cards.
Ref:
- https://superuser.com/questions/542740/linux-kernel-scheduler-for-a-pen-drive-noop-or-deadline
- http://forum.xda-developers.com/showpost.php?s=b236833a23a9d84f6f1b863c120ccda7&p=23616564&postcount=4
.

I've been using this setting for several months on different Pi's with no problems.

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -34,7 +34,7 @@ ip_nameservers=
 drivers_to_load=
 online_config=
 usbroot=
-cmdline="dwc_otg.lpm_enable=0 console=ttyAMA0,115200 kgdboc=ttyAMA0,115200 console=tty1 elevator=deadline"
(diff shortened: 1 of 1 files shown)


goranche:
uhm... :thought_balloon: I disagree with this one and would leave it set to deadline...

yes, noop can be better in certain scenarios, but not in most... deadline has better problem situation recovery, in particular on systems that do parallel I/O (databases, file server, ...)
and of course, I/O performance will be better with deadline (also mentioned on the pages you linked)

the default is "good enough" (in this case I'd argue better than the change) and is something individuals should set according to their needs...

diederikdehaas:
https://github.com/debian-pi/raspbian-ua-netinst/pull/152 :wink:

Issue #383, opened 2016-03-07
duplicate localhost entry in /etc/hosts


richard-scott:
I've noticed that in the latest 1.1.x release, that localhost is duplicated in /etc/hosts:
```
root@pi:~# cat /etc/hosts
127.0.0.1 localhost
127.0.0.1       localhost
127.0.1.1       pi
root@pi:~#
```
I can only guess that when [this](https://github.com/debian-pi/raspbian-ua-netinst/blob/v1.1.x/scripts/etc/init.d/rcS#L769-L777) runs that there is already an entry for localhost in /etc/hosts.

Maybe line 777 needs altering to overwrite /etc/hosts instead of appending to it.

diederikdehaas:
Are you installing `wheezy` or `jessie` (if you didn't specify it, it would be `jessie`)?

richard-scott:
I ran this today:

```
git clone -b v1.1.x --single-branch https://github.com/debian-pi/raspbian-ua-netinst.git
cd raspbian-ua-netinst
./clean.sh
./update.sh
./build.sh
./buildroot.sh
dd if=raspbian-ua-netinst-20160307-gitf0a95bc.img of=/dev/sdc
```

Did not edit any files, so used defaults where applicable.

Mausy5043:
Confirmed on `jessie` Pi1B and P1B+.

Mausy5043:
I wouldn't worry too much about this. There is no negative impact on the system. It just looks less perfect.
Nevertheless, a good find. :+1:

Mausy5043:
Personally, I always add these:

```
echo "::1             localhost ip6-localhost ip6-loopback" >> /rootfs/etc/hosts
echo "fe00::0         ip6-localnet" >> /rootfs/etc/hosts
echo "ff00::0         ip6-mcastprefix" >> /rootfs/etc/hosts
echo "ff02::1         ip6-allnodes" >> /rootfs/etc/hosts
echo "ff02::2         ip6-allrouters" >> /rootfs/etc/hosts
```

diederikdehaas:
Hmm ... I really wonder how this came out to be, just like the double `lo` definition in `/etc/network/interfaces` on `wheezy`.

Mausy5043:
I'm guessing someone changed something upstream.

goranche:
I've added a check for an existing `localhost` in `/rootfs/etc/hosts` to #381

richard-scott:
[...]

PR #385, opened 2016-03-11
Single loopback for v1.1.x

goranche:
Pretty much the same thing as #381, just for the v1.1.x branch
one thing that's added is a check (and removal) for 127.0.1.1, as the script sets this later on.


goranche:
This fixes #383

diederikdehaas:
> This fixes #383

AFAIK it needs to be part of the commit message :wink:

goranche:
nah, not the commit message, but the description... but I thought it worked in the comments as well...
might have been gitlab

PR #388, opened 2016-03-18
Fix hosts file creation

### LINKED ISSUE #387: /etc/hostname blank after install
After building the installer from  the v1.1.x branch I have an empty /etc/hostname file after rebooting., it 

Steps to reproduce:

1. git clone -b v1.1.x --single-branch https://github.com/debian-pi/raspbian-ua-netinst.git
2. build the latest image downloaded
3. dd'd a new image onto my SD card, made no other changes and booted it.

My /installer-config.txt does not exist as nothing has been changed from defaults.

goranche:
It adds a newline to `/etc/hostname` (same will be done for v1.0.x)
In addition to this, `/etc/hosts` gets generated in a different way and afterwards contains needed entries

Fixes #387

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -789,22 +789,26 @@ echo "OK"
 
 # default hostname
 echo -n "  Configuring hostname... "
-echo -n $hostname > /rootfs/etc/hostname || fail
+echo $hostname > /rootfs/etc/hostname || fail
+echo "OK"
+
+echo -n "  Configuring hosts... "
 touch /rootfs/etc/hosts
-{
-    # Add localhost to hosts (if needed)
-    if ! grep -q "localhost" /rootfs/etc/hosts; then
-        echo "127.0.0.1	localhost"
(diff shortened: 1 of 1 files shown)


Mausy5043:
I'd love an explanation why this works.

Mausy5043:
See my previous comment above, could you elaborate on the root-cause? (no challenging your solution) Just me not understanding how the `{ }` construct could cause this problem.

goranche:
the script is run by (a quite limited) ash from the busybox package... combined with a quite complicated setup, including redirected standard output, and using blocks that are then output to a file...
it should be straight forward recreating the conditions that lead to this situation, and go from there...

replacing the code with a simpler version that doesn't rely on blocks and/or subshells is just a safer option... and works

> See my previous comment above

yes, I got it the first time... :smile:  7 minutes in between, hold your horses :wink:

Mausy5043:
I'm merging because it works. 

Still don't understand fully why it didn't before for some people.

PR #389, opened 2016-03-18
Add trailing newline to /etc/hostname

goranche:
This one is related to #388 but fixed the hostname issue for v1.0.x
It adds a trailing newline (well, just removed the option that prevented it from being created) to `/etc/hostname`

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -527,7 +527,7 @@ echo "OK"
 
 # default hostname
 echo -n "  Configuring hostname... "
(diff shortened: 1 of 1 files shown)


Mausy5043:
@goranche Could you explain this? How does the `-n` option prevent the file from being created?

goranche:
> How does the -n option prevent the file from being created?

huh? who said it does?  :confused:   it just prevents a newline to be added to the end of the line

Mausy5043:
Ah OK

PR #390, opened 2016-03-18
Add watchdog functionality

### LINKED ISSUE #298: Add option for watchdog service
Apparently the Pi has a hardware watchdog timer, see http://blog.ricardoarturocabral.com/2013/01/auto-reboot-hung-raspberry-pi-using-on.html

It could be a nice feature to add.

goranche:
Fixes #298

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -39,6 +39,7 @@ rootfstype=ext4
 final_action=reboot
 hardware_versions=detect
 hwrng_support=1
+enable_watchdog=1
 gpu_mem=
 
 # these shouldn't really be changed unless auto-detection fails
@@ -1007,6 +1008,12 @@ fi
 echo "snd-bcm2835" >> /rootfs/etc/modules
 if [ "$hwrng_support" = "1" ]; then
     echo "bcm2708-rng" >> /rootfs/etc/modules
+fi
+
+# (conditionaly) enable hardware watchdog and set up systemd to use it
+if [ "${enable_watchdog}" = "1" ]; then
+    echo "bcm2708_wdog" >> /rootfs/etc/modules
+    sed -i 's/^.*RuntimeWatchdogSec=.*$/RuntimeWatchdogSec=20s/' /rootfs/etc/systemd/system.conf
 fi
 
 echo "Configuring bootloader to start the installed system..."
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -40,7 +40,7 @@ rootfstype=ext4
 final_action=reboot
 hardware_versions=detect
 hwrng_support=1
-enable_watchdog=1
+enable_watchdog=0
 gpu_mem=
 
 # these shouldn't really be changed unless auto-detection fails
@@ -1025,7 +1025,7 @@ fi
 # (conditionaly) enable hardware watchdog and set up systemd to use it
 if [ "${enable_watchdog}" = "1" ]; then
     echo "bcm2708_wdog" >> /rootfs/etc/modules
-    sed -i 's/^.*RuntimeWatchdogSec=.*$/RuntimeWatchdogSec=20s/' /rootfs/etc/systemd/system.conf
+    sed -i 's/^.*RuntimeWatchdogSec=.*$/RuntimeWatchdogSec=14s/' /rootfs/etc/systemd/system.conf
 fi
 
 echo "Configuring bootloader to start the installed system..."
--- a/README.md
+++ b/README.md
@@ -161,8 +161,8 @@ The format of the _installer-config.txt_ file and the current defaults:
                               # enabled (1) on all presets. Users requiring a `base` install are advised that
                               # `hwrng_support=0` must be added in `installer-config.txt` if HWRNG support is
                               # undesirable.
-    enable_watchdog=1         # loads up the hardware watchdog module and configures systemd to use it. Set to
-                              # anything else but "1" to disable this functionality.
+    enable_watchdog=0         # loads up the hardware watchdog module and configures systemd to use it. Set to
+                              # "1" to enable this functionality.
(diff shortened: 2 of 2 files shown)


Mausy5043:
Having had some bad experiences with a watchdog ending up rebooting my device in a loop. I'd love the settings to be a bit more relaxed. Would you consider setting the timerdefault at one or two minutes?
This gives the user time to try and fix things via a terminal.

goranche:
> setting the timerdefault at one or two minutes

don't think that should be needed, no... systemd should take care of making sure the hardware watchdog doesn't restart the system... as it's configured in this PR, it will "ping" the watchdog every 10 seconds, so it shouldn't be an issue, except if it fails to do so, in which case you'd be out of luck anyway...

Mausy5043:
I've tested this and it seems to work (module is loaded and setting is inserted correctly). Anybody have a "simple" way of blocking out the kernel for >20s to force the watchdog to cause a reboot? 

I was trying to `git clone https://github.com/raspberrypi/linux.git` and that triggered a reboot. Not sure if it was related to the watchdog though. Did see some `oom-killer`s as it progressed...

goranche:
> blocking out the kernel

not so much the kernel as systemd pinging of the watchdog device...

Mausy5043:
Whatever, as long as it achieves the goal.

Device is up:
```
$ la /dev/w*
crw------- 1 root root 10, 130 Mar 20 18:17 /dev/watchdog
```

goranche:
> Whatever

:smirk: just thought you'd like to understand how the watchdog actually functions...

anyway... not sure what you're trying to test... you can disable systemds watchdog handling, and then do a simple `cat > /dev/watchdog`... as long as you keep sending characters (don't use `V` :thought_balloon: ) into the device, the hardware watchdog timer is reset... when you stop doing that, the device will sooner or later (about 10 seconds in my case) reboot...

goranche:
> don't use V

never mind... it seems the `bcm2708_wdog` doesn't support the "Magic Close" functionality... :smile: 
if you have time to spare, here's some [light reading](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/watchdog/watchdog-api.txt)...

Mausy5043:
I kinda know how it works. Was just wondering if there's a way to starve the watchdog.

goranche:
> I was trying to `git clone ...` and that triggered a reboot

I'm a bit confused / worried by this one, though... :thought_balloon:

Mausy5043:
Yes, I need to do some extra testing to make sure it's unrelated.

Mausy5043:
This log-entry bothers me:

```
Mar 21 16:17:02 rbian systemd[1]: Started udev Coldplug all Devices.
Mar 21 16:17:02 rbian kernel: bcm2708 watchdog, heartbeat=10 sec (nowayout=0)
Mar 21 16:17:02 rbian systemd[1]: Hardware watchdog 'BCM2708', version 1
Mar 21 16:17:02 rbian systemd[1]: Failed to set timeout to 20s: Invalid argument
```

Mausy5043:
This causes crashes
```
$ mkdir tmp
$ cd tmp
$ git clone -b "rpi-4.3.y" --single-branch https://github.com/raspberrypi/linux.git
```
I just tested this on a non-watchdog Pi and `oom-killer` said (among others):
```
Mar 21 18:05:50 rbups kernel: Out of memory: Kill process 1451 (git) score 849 or sacrifice child
Mar 21 18:05:50 rbups kernel: Killed process 1451 (git) total-vm:471580kB, anon-rss:418604kB, file-rss:176kB
```
The Pi with watchdog activated just reboots.

diederikdehaas:
With what Pi version and how much RAM does it have?

goranche:
> causes crashes

what's a "crash" in this context?

Mausy5043:
Both are Pi1B+; 482MB total memory (GPU = 16MB)
Crash is defined here as: uncleanly rebooting where I expect the behaviour to be that `oom-killer` jumps in and saves the day.

The reboot looks more like a power-cycle than a clean `shutdown -r`. The logs abruptly stop.

goranche:
uhm... I think you misunderstand what the watchdog does here... :thought_balloon: 

it resets the chip... that's it, there is no clean shutdown, there is nothing to save the day...

Mausy5043:
Well, then I don't like the `watchdog`. `oom-killer` is much nicer. :wink:

goranche:
the two have nothing in common, really, they solve completely different issues in different ways

oom-killer will not save you if the kernel hangs (panics)... it will not make sure a reboot will actually happen, even if something goes wrong, ... the hardware watchdog, however, can do those things

Mausy5043:
Yes, but in this particular case it would have been nice if the watchdog would have waited a bit longer.
I don't know. Tired... need... sleep... :zzz:

diederikdehaas:
> This log-entry bothers me
https://github.com/debian-pi/raspbian-ua-netinst/issues/298#issuecomment-167235672
IOW: _I found that the BCM2708 watchdog has a maximum 15 second timeout._

scripts/etc/init.d/rcS
```diff
@@ -39,6 +39,7 @@ rootfstype=ext4
 final_action=reboot
 hardware_versions=detect
 hwrng_support=1
+enable_watchdog=1
```
Mausy5043: I'd prefer this to read `enable_watchdog=0`.

diederikdehaas: Me too

goranche:
> the BCM2708 watchdog has a maximum 15 second timeout

:unamused: sheesh... I'll change systemds timeout to 14 seconds then...

scripts/etc/init.d/rcS
goranche: agree... will change

goranche:
> to 14 seconds

why 14? excellent question, I've asked that myself as well... :innocent:

PR #391, opened 2016-03-18
Add option to disable predictable names

### LINKED ISSUE #103: Switching board causes network not to work anymore
Yesterday I installed a new image on a pi. Network works as expected.

After that I put the sd-card in another pi and eth0 was not working anymore, so no network access was possible.

Searching in the forums resulted in two post that describe (and fix) this problem:

http://www.raspberrypi.org/forums/viewtopic.php?f=28&t=82467
http://www.raspberrypi.org/forums/viewtopic.php?f=6&t=5442

It turns out that the /etc/udev/rules.d/70-persistent-net.rules added a new rules where eth1 is linked to the mac of the second pi and therefore a mismatch with /etc/network/interfaces

Moving the rules-file out of the way causes that file to be recreated again on reboot with the proper mac address.

In my original installation (based on an official foundation image from the early days) this file is not created at all. Switching boards always worked. I haven't tried a recent image.

With this behavior (adding eth1 to the new mac) switching a sd card between different pi's is sort of impossible.

goranche:
Fixes #103

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -837,8 +837,9 @@ touch /rootfs/etc/network/interfaces || fail
 
 if [ "${disable_predictable_nin}" = "1" ]; then
     # as described here: https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames
-    # adding net.ifnames=0 to /boot/cmdline
+    # adding net.ifnames=0 to /boot/cmdline and disabling the persistent-net-generator.rules
     cmdline="${cmdline} net.ifnames=0"
+    ln -s /dev/null /rootfs/etc/udev/rules.d/75-persistent-net-generator.rules
 fi
 
 if [ "$ip_addr" != "dhcp" ]; then
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -838,7 +838,7 @@ touch /rootfs/etc/network/interfaces || fail
 if [ "${disable_predictable_nin}" = "1" ]; then
     # as described here: https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames
     # adding net.ifnames=0 to /boot/cmdline
-    sed -i 's/^\(.*\)$/\1 net.ifnames=0/' /bootfs/cmdline
+    cmdline="${cmdline} net.ifnames=0"
 fi
 
 if [ "$ip_addr" != "dhcp" ]; then
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -25,6 +25,7 @@ timeserver=time.nist.gov
 timezone=Etc/UTC
 locales=
 system_default_locale=
+disable_predictable_nin=1
 ifname=eth0
 ip_addr=dhcp
 ip_netmask=0.0.0.0
@@ -833,6 +834,11 @@ touch /rootfs/etc/network/interfaces || fail
         echo "wpa-conf /boot/config/wpa_supplicant.conf"
     fi
 } >> /rootfs/etc/network/interfaces || fail
+
+if [ "${disable_predictable_nin}" = "1" ]; then
+    # as described here: https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames
+    ln -s /dev/null /rootfs/etc/udev/rules.d/80-net-setup-link.rules
+fi
 
 if [ "$ip_addr" != "dhcp" ]; then
     cp /etc/resolv.conf /rootfs/etc/ || fail
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -837,7 +837,8 @@ touch /rootfs/etc/network/interfaces || fail
 
 if [ "${disable_predictable_nin}" = "1" ]; then
     # as described here: https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames
-    ln -s /dev/null /rootfs/etc/udev/rules.d/80-net-setup-link.rules
+    # adding net.ifnames=0 to /boot/cmdline
+    sed -i 's/^\(.*\)$/\1 net.ifnames=0/' /bootfs/cmdline
 fi
 
 if [ "$ip_addr" != "dhcp" ]; then
--- a/README.md
+++ b/README.md
@@ -136,6 +136,10 @@ The format of the _installer-config.txt_ file and the current defaults:
     locales=                  # a space delimited list of locales that will be generated during install
                               # (e.g. "en_US.UTF-8 nl_NL sl_SI.UTF-8")
     system_default_locale=    # the default system locale to set (using the LANG environment variable)
+    disable_predictable_nin=1 # Disable Predictable Network Interface Names. Set to 0 if you want to use predictable
+                              # network interface names, which means if you use the same SD card on a different
+                              # RPi board, your network device might be named differently. This will result in the
+                              # board having no network connectivity.
     ifname=eth0
     ip_addr=dhcp
     ip_netmask=0.0.0.0


goranche:
this one needs some testing... it works for me, but on a single board (with a second for testing)...
would be nice if others can confirm it works as intended

richard-scott:
This doesn't work for me as my udev file has a different name:

```
root@rpi2:/etc/udev/rules.d# cat /etc/udev/rules.d/70-persistent-net.rules
# This file was automatically generated by the /lib/udev/write_net_rules
# program, run by the persistent-net-generator.rules rules file.
#
# You can modify it, as long as you keep each rule on a single
# line, and change only the value of the NAME= key.

# USB device 0x:0x (smsc95xx)
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{address}=="b8:27:eb:18:91:48", ATTR{dev_id}=="0x0", ATTR{type}=="1", KERNEL=="eth*", NAME="eth0"
root@rpi2:/etc/udev/rules.d#
```

Also, this is standard udev functionality for any Linux OS so it wise to disable this by default? I would suggest that it is nicer for the default to be "disable_predictable_nin=0" and if you want to activate this option you then manually enable it at install time?

Being perfectly honest. I don't see a need for this at all in the installer as only a few will need it, and you can do this via the 'post-install.txt' file.

goranche:
on a "standard" OS/computer install you're unlikely to change the drive you installed to, to another computer...
on a RPi, however, changing (or copying) SD cards to different RPi devices is common... what isn't, though, is adding / removing network hardware in a manner that would break network device naming... :thought_balloon: 
considering this, IMHO disabling predictable device naming should be enabled by default...

> and you can do this via the 'post-install.txt' file

you can do a full system install from post-install.txt, including system bootstrap... :wink:

goranche:
> my udev file has a different name

not sure I understand... what you showed is just one of the possible rules files in `/etc/udev/rules.d` :thought_balloon:

Mausy5043:
I think what @richard-scott is trying to say is that in his install (and mine) the MAC is stored in `/etc/udev/rules.d/70-persistent-net.rules` whereas your fix is to symlink `/etc/udev/rules.d/80-net-setup-link.rules`. The latter doesn't exist on mine (and I'm guessing on @richard-scott's).

However, in your defense, @goranche, the webpage you refer to in your PR ([this one](https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/)) suggests the exact fix that you propose. So, going to need to test this.

Finally, I was going to argue the default value for `disable_predictable_nin`... But the value `1` will yield a working system for everybody and a value of `0` would mean some users finding their Pi not connecting when switching SD-cards between Pi's. So, I'm not going to argue this.

Mausy5043:
After booting on the device that was installed on:
```
$ la /etc/udev/rules.d/
total 12
drwxr-xr-x 2 root root 4096 Mar 19 11:55 .
drwxr-xr-x 4 root root 4096 Mar 19 11:16 ..
-rw-r--r-- 1 root root  436 Mar 19 11:55 70-persistent-net.rules
lrwxrwxrwx 1 root root    9 Mar 19 11:50 80-net-setup-link.rules -> /dev/null
```
```
$ ifconfig
eth0      Link encap:Ethernet  HWaddr b8:27:eb:2f:84:06
          inet addr:10.0.1.11  Bcast:10.0.1.255  Mask:255.255.255.0
          inet6 addr: fe80::ba27:ebff:fe2f:8406/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:2937 errors:0 dropped:0 overruns:0 frame:0
          TX packets:828 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:262859 (256.6 KiB)  TX bytes:195108 (190.5 KiB)

lo        Link encap:Local Loopback
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)
```

Mausy5043:
Booting with this SD-card in another Pi results in an "island-mode" Pi.
In `dmesg` it says:
`systemd-udevd:  renamed network interface eth0 to eth1`

`ifconfig` only shows the `lo` interface.
:disappointed:

diederikdehaas:
My guess is that on Fedora it is 80... (LP is a RedHat/Fedora guy)

goranche:
@diederikdehaas nope... systemd has changed the way predictable naming is handled throughout versions (I'll act surprised if it helps)...

first (udev 171 - 208) `/lib/udev/rules.d/80-net-name-slot.rules` was used to hold rules for naming...
then (udev 208 -216) `/lib/udev/rules.d/80-net-setup-link.rules`was used instead...
now `80-net-setup-link.rules` is just a trigger for `/lib/systemd/network/99-default.link` (which can be overridden in `/etc/systemd/network/99-default.link`)...

ask me again why I don't really like systemd...

oh, the files in `/lib/udev/rules.d/` can be overridden in `/etc/udev/rules.d/`, in case you're wondering...
yes, the "configuration" gets stored into `70-something.rules` (don't care enough to look it up), but that's not the solution... the solution is to stop the naming at all...

goranche:
I've pushed a new change... this one adds `net.ifnames=0` to `/boot/cmdline`... supposedly this should work...
can't try it out at the moment... didn't take a raspberry device when I went scuba diving for the day :sunglasses:

goranche:
oh, and the link to `/dev/null` is removed... will test this one in the evening (I hope)

diederikdehaas:
Why not (something like) the solution presented at https://github.com/debian-pi/raspbian-ua-netinst/issues/103#issuecomment-58824684 ?
I thought that was rather neat and also what the RPF uses in their image?

goranche:
because that changes files that come with the package...

scripts/etc/init.d/rcS
```diff
@@ -834,6 +835,12 @@ touch /rootfs/etc/network/interfaces || fail
     fi
 } >> /rootfs/etc/network/interfaces || fail
 
+if [ "${disable_predictable_nin}" = "1" ]; then
+    # as described here: https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames
+    # adding net.ifnames=0 to /boot/cmdline
+    sed -i 's/^\(.*\)$/\1 net.ifnames=0/' /bootfs/cmdline
```
Mausy5043: `/bootfs/cmdline.txt`

Mausy5043: Also [this line](https://github.com/goranche/raspbian-ua-netinst/blob/unpredictable-interfaces/scripts/etc/init.d/rcS#L1073) isn't going to help either.

goranche: good catch on both of them :+1:

scripts/etc/init.d/rcS
```diff
@@ -834,6 +835,12 @@ touch /rootfs/etc/network/interfaces || fail
     fi
 } >> /rootfs/etc/network/interfaces || fail
 
+if [ "${disable_predictable_nin}" = "1" ]; then
+    # as described here: https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames
+    # adding net.ifnames=0 to /boot/cmdline
+    cmdline="${cmdline} net.ifnames=0"
```
Mausy5043: not working for me :disappointed:

Mausy5043: However, I've found that this:
```
sudo ln -s /dev/null /etc/udev/rules.d/75-persistent-net-generator.rules
```
solves the problem.

Source: https://askubuntu.com/questions/628217/use-of-predictable-network-interface-names-with-alternate-kernels  -- at the bottom of the accepted answer.

goranche: hm... could you double check `net.ifnames=0` is actually added to `cmdline.txt`?

goranche: > I've found that this

LOL   another change? :grin:

Mausy5043: Yup.

Mausy5043: I've tested with manually adding `net.ifnames=0`  --> no luck  
and with manually adding `net.ifnames=0 biosdevname=0` --> nope

goranche: I'll change it a bit later, and try it out... I'll let you know when I commit it...

goranche: > no luck

yeah... it's nuts... obvious it doesn't behave as documented... oh well...

goranche: the thing I really don't like is that in the end we'll be deleting the stored names at every boot...
that's just ... "jucky"

Mausy5043: PR #380 was very helpfull :+1:

Mausy5043: We could think about whether this should be an opt-in or an opt-out feature. :thought_balloon:

goranche: > whether this should be an opt-in or an opt-out feature

we could, I guess... I (obviously) think it should be opt-out feature, because of the specifics of the device and likely use cases...

were this a desktop computer, where you can switch network interfaces and who know what, sure, I'd go with the opt-in, but being what it is, however...

goranche: > PR #380 was very helpfull

glad to hear :+1: I like it a lot myself :wink:

I'm actually still abusing it (with some jumping through hoops... oh well) to hook into the installer after fdisk is done, so I can set up an additional partition without the need for a separate reboot...

Mausy5043: I tend to agree with this being an opt-out.

goranche: not much to discuss then :grin: 

also, converting it to opt-in is as simple as setting [`disable_predictable_nin`](https://github.com/debian-pi/raspbian-ua-netinst/pull/391/files#diff-ef22c8fabe7f21fa3c3ab45495d6774bR28) to 0

diederikdehaas: > net.ifnames=0

IIRC that option was added to systemd/udev in a version later then is included in `jessie`. It used to be mentioned in that wiki, but I guess they _assumed_ that everyone would have that version by now. Maybe webarchive has an old version if you really curious.

goranche: > in a version later

huh... would explain a lot, I guess... :thought_balloon:

goranche: > IIRC that option was added to systemd/udev in a version later then is included in jessie

nah... doesn't seem like it, the version that gets installed is 215, and the kernel command line parameter is supposed to be present way before that...

goranche: @Mausy5043  I've added creating a soft link to `/dev/null` for `/etc/udev/rules.d/75-persistent-net-generator.rules`, and it seems to work for me...

Mausy5043: I assume you kept the `net.ifnames=0` for future kernels?

goranche: for some reason I'm hoping it will work someday... hope dies last and all that...

Mausy5043:
This should fix #103

goranche:
I might smash my raspberries, if it doesn't :unamused:

Mausy5043:
raspberry jam :-D

goranche:
with crunchy bits... :+1:

PR #393, opened 2016-03-23
No sound

Mausy5043:
Not loading the module `snd-bcm2835` by default.
Minimal systems do not require this module.

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -1016,8 +1016,6 @@ if [ "$ifname" != "eth0" ]; then
     fi
 fi
 
-# add reasonable default modules, works only after kernel is properly installed
-echo "snd-bcm2835" >> /rootfs/etc/modules
 if [ "$hwrng_support" = "1" ]; then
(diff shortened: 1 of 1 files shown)


diederikdehaas:
I'm completely for this :+1:

bbinet:
Why not make it configurable? sound might be important for some applications.

Mausy5043:
IMHO, that point could be made for just about any module or package.

In fact this **is** configurable.

The installer provides the end-user with a target system that (we think) will be a good basis to build their own custom system on top of. We want to keep this as minimal as possible to make it appealing to as many users as possible so we have to make choices what to include and what not. If we feel that many users may want to install a given option then we provide either an opt-in (e.g. watchdog) or opt-out (e.g. HWRNG support).
The `installer-config.txt` allows for the installation of additional packages (e.g. nut) that are not provided by default by the installer.
The `post-install.txt` provides the end-user with a mechanism of adding specific items (e.g. sound support) to the target system.

bbinet:
Ok, configuring it through `post-install.txt` is fine.

Issue #349, opened 2015-12-29
wpa_supplicant.conf is stored at an odd location


diederikdehaas:
Currently `scripts/etc/init.d/rcS` has the following code block (line 776-778):
```
if [ "$ifname" != "eth0" ]; then
    echo "wpa-conf /boot/config/wpa_supplicant.conf" >> /rootfs/etc/network/interfaces || fail
fi
```
While it is correct and should work, it is a rather odd location for the `wpa_supplicant.conf` file

Mausy5043:
If you want to start moving code around, think twice. No, think three times. You might be opening a boar's pit :grinning: .  
You're probably aware that there are internal dependencies. This could take a lot of effort.

Just saying.

diederikdehaas:
Location of the file. `/boot/config/wpa_supplicant.conf` isn't the 'normal' location, that would be `/etc/wpa_supplicant/wpa_supplicant.conf`

diederikdehaas:
A general thought (of mine):
_If you're afraid to change code because it might brake something, your code is already broken. 
It just hasn't fallen apart **yet**_

Mausy5043:
:+1:

goranche:
it would be better if we added a `/config/interfaces.d` directory, and copied over all files found in that location to `/rootfs/etc/network/interfaces.d`
and yes, I agree that wpa_supplicant.conf should **never** be copies over `/etc/network/interfaces` as the two are different beasts...

it's not so much the location in `/config` that worries me, but where it get's copied to (overwriting the interfaces file)

Mausy5043:
How about : 
```
/bootfs
+ config
  + apt
  + installer
  - overlay
    - etc
      - wpa_supplicant.conf
```
Copy the contents of `/bootfs/config/overlay` to `/rootfs`. (I think this was discussed before, although not my method of choice).
`wpa_supplicant.conf` tends to be installed in `/etc` together with `/etc/wpa_supplicant`.

diederikdehaas:
> copied over all files found in that location to /rootfs/etc/network/interfaces.d

Never thought about putting it there as it's also a non-standard location, but it's certainly not unlogical.

> wpa_supplicant.conf should never be copies over /etc/network/interfaces

Correct, but `/etc/network/interfaces` should contain a reference/link to the wpa_supplicant.conf, which is what the quoted code does :wink: 

>  wpa_supplicant.conf tends to be installed in /etc together with /etc/wpa_supplicant.

[...]

PR #394, opened 2016-03-23
Overlay target system

Mausy5043:
Allow end-user to supply custom files.
User supplied files are added to the target system during install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,8 @@ To have the installer actually copy the files to the target system, add the foll
 ```
 install_files my-files.list
 ```
-  where `my-files.list` is the name of the file containing the list of files.
+where `my-files.list` is the name of the file containing the list of files.
+If needed, you can call `install_files` multiple times with different list files.  
 Please be aware that some restrictions may apply to the sum of the file sizes. If you wish to supply large files in this manner you may need to adjust the value of the `bootsize` parameter.
 
 ### Custom installer script
--- a/README.md
+++ b/README.md
@@ -174,6 +174,14 @@ Presets set the `cdebootstrap_cmdline` variable. For example, the current _serve
 
 (If you build your own installer, which most won't need to, and the configuration files exist in the same directory as this `README.md`, it will be include in the installer image automatically.)
 
+### Bring your own files
+You can have the installer place your custom configuration files (or any other file you wish to add) on the installed system during the installation. For this, you need to provide the necessary files in the `/config/files/` directory of your SD card (you may need to create this directory if it doesn't exist). The `/config/files/` directory is the root-point. It must have the same structure as inside the installed system. So, a file that you place on the SD card in `/config/files/etc/wpa_supplicant/wpa_supplicant.conf` will end up on the installed system as `/etc/wpa_supplicant/wpa_supplicant.conf`.
+Each file or directory that you wish to place on the target system must also be listed in a configuration file in the directory `/config` on your SD card. This allows you to specify the owner (and group) and the permissions of the file. An example file is provided with the installer (see `/config/my-files.list` for more information). ONLY files listed there are copied over to the installed system.
+To have the installer actually copy the files to the target system, add the following command at an appropriate point in your `post-install.txt` file:
+`install_files my-files.list`
+  where `my-files.list` is the name of the file containing the list of files.
+Please be aware that some restrictions may apply to the sum of the file sizes. If you wish to supply large files in this manner you may need to adjust the value of the `bootsize` parameter.
+
 ### Custom installer script
 
 It is possible to replace the installer script completely, without rebuilding the installer image. To do this, place a custom `rcS` file in the config directory of your SD card. The installer script will check this location and run this script instead of itself. Take great care when doing this, as it is intended to be used for development purposes.
--- a/README.md
+++ b/README.md
@@ -178,7 +178,9 @@ Presets set the `cdebootstrap_cmdline` variable. For example, the current _serve
 You can have the installer place your custom configuration files (or any other file you wish to add) on the installed system during the installation. For this, you need to provide the necessary files in the `/config/files/` directory of your SD card (you may need to create this directory if it doesn't exist). The `/config/files/` directory is the root-point. It must have the same structure as inside the installed system. So, a file that you place on the SD card in `/config/files/etc/wpa_supplicant/wpa_supplicant.conf` will end up on the installed system as `/etc/wpa_supplicant/wpa_supplicant.conf`.
 Each file or directory that you wish to place on the target system must also be listed in a configuration file in the directory `/config` on your SD card. This allows you to specify the owner (and group) and the permissions of the file. An example file is provided with the installer (see `/config/my-files.list` for more information). ONLY files listed there are copied over to the installed system.
 To have the installer actually copy the files to the target system, add the following command at an appropriate point in your `post-install.txt` file:
-`install_files my-files.list`
+```
+install_files my-files.list
+```
   where `my-files.list` is the name of the file containing the list of files.
 Please be aware that some restrictions may apply to the sum of the file sizes. If you wish to supply large files in this manner you may need to adjust the value of the `bootsize` parameter.
 
--- a/config/my-files.list
+++ b/config/my-files.list
@@ -2,7 +2,7 @@
 # Add one line per file using this format: 
 # <ownership> <permission> <path/filename>
 #
-# <ownership>     [OWNER][:[GROUP]] (see `man chown`)
+# <ownership>     OWNER[:GROUP] (see `man chown`)
 # <permission>    MODE  (see `man chmod`)
 # <path/file>     /path/as/it/will/appear/on/the/target/system.txt
 # 
--- a/config/my-files.list
+++ b/config/my-files.list
@@ -1,0 +1,14 @@
+# To add custom files on the target system:
+# Add one line per file using this format: 
+# <ownership> <permission> <path/filename>
+#
+# <ownership>     [OWNER][:[GROUP]] (see `man chown`)
+# <permission>    MODE  (see `man chmod`)
+# <path/file>     /path/as/it/will/appear/on/the/target/system.txt
+# 
+# Source files are stored in the directory `config/files` in the same path as
+# where they will be stored on the target system.
+#
+# EXAMPLE:
+# A file placed on the SD-card in /config/files/home/pi/.bashrc is listed as follows:
+# pi:pi  755  /home/pi/.bashrc
--- a/config/my-files.list
+++ b/config/my-files.list
@@ -12,3 +12,8 @@
 # EXAMPLE:
 # A file placed on the SD-card in /config/files/home/pi/.bashrc is listed as follows:
 # pi:pi  755  /home/pi/.bashrc
+#
+# Don't forget to call the function:
+# install_files my-files.list
+# from your `post-install.txt` where "my-files.list" is this file. 
+# The function `install_files` will assume the file is located in the `config` directory.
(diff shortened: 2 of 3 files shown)


scripts/etc/init.d/rcS
```diff
@@ -85,6 +85,24 @@ sanitize_inputfile()
     fi
 }
 
+install_files()
+{
+  file_to_read=$1
+  sanitize_inputfile file_to_read
+  while read -r line; do
+    comment=$(echo "$line" | cut -c1)
```
goranche: I think you should be able to replace the comment filtering with the following construct:
```
while read -r line; do
        <do stuff with line>
done < "${file_to_read}" | grep -v ^#
```

config/my-files.list
```diff
@@ -0,0 +1,14 @@
+# To add custom files on the target system:
+# Add one line per file using this format: 
+# <ownership> <permission> <path/filename>
+#
+# <ownership>     [OWNER][:[GROUP]] (see `man chown`)
```
goranche: ownership should be `owner[:group]`

goranche: unless you want to make owner optional :innocent:

scripts/etc/init.d/rcS
```diff
@@ -85,6 +85,24 @@ sanitize_inputfile()
...
+    if [ ! "$comment" = "#" ]; then               # lines starting with # get ignored
+      owner=$(echo "$line" | awk '{ print $1 }')
+      perms=$(echo "$line" | awk '{ print $2 }')
+      file=$(echo  "$line" | awk '{ print $3 }')
+      mkdir -p ../rootfs"$(dirname "$file")"
```
goranche: why the `..`?
and please add the `/rootfs` inside the quoted string, like so: `"/rootfs/$(dirname "${file}")"`

goranche: same goes for next two lines

scripts/etc/init.d/rcS
```diff
@@ -85,6 +85,24 @@ sanitize_inputfile()
...
+      cp -v config/files"$file" ../rootfs"$file"
```
goranche: could you add a check if "config/files${file}" is a directory?
that way we'd be able to create and (even more important) set permissions on directories as well...

all else should stay the same, just skip the copy part if "config/files${file}" is a directory :thought_balloon:

goranche:
good stuff :+1:   I've commented on some minor things, but otherwise, I like it, thanks :wink:

scripts/etc/init.d/rcS
Mausy5043: I was struggling with that too. I'll take a look.

scripts/etc/init.d/rcS
Mausy5043: Oops. That's left over from the testing. Good catch! :+1:

scripts/etc/init.d/rcS
Mausy5043: not in my version of `busybox` :-(

Mausy5043: The `<` redirects the file contents to the `while` loop. Any output to `stdout` from the `while` loop (notably `cp -rv` output) gets piped to `grep`.

Also `done < grep "-v "^#" "$file_to_read"` doesn't work.

goranche: > output to stdout from the while loop gets piped to grep

should not happen, no... :thought_balloon:

scripts/etc/init.d/rcS
```diff
@@ -96,7 +96,9 @@ install_files()
       perms=$(echo "$line" | awk '{ print $2 }')
       file=$(echo  "$line" | awk '{ print $3 }')
       mkdir -p "/rootfs$(dirname "$file")"
-      cp -v "config/files$file" "/rootfs$file"
+      if [ ! -d "config/files$file" ]; then
```
Mausy5043: Don't think `dirname` will recognise it when `$file` is really a dir

scripts/etc/init.d/rcS
goranche: yeah... my bad... :blush: 

this will (tested with busybox static):
```
grep -v ^# "${file_to_read}" | while read -r line; do
        <do stuff with line>
done
```

Mausy5043: brilliant

scripts/etc/init.d/rcS
```diff
@@ -85,6 +85,25 @@ sanitize_inputfile()
     fi
 }
 
+install_files()
+{
+  file_to_read=$1
+  sanitize_inputfile file_to_read
+  grep -v "^#\|^$" "${file_to_read}" | while read -r line; do
```
Mausy5043: Now also excluding empty lines.

goranche: good idea... you could also go a bit further and use this regex:
`^[[:space:]]*#\|^[[:space:]]*$`

this will allow whitespace (just whitespace) in front of the hash sign... and it'll match lines with just whitespace as well...

scripts/etc/init.d/rcS
```diff
@@ -85,6 +85,25 @@ sanitize_inputfile()
     fi
 }
 
+install_files()
+{
+  file_to_read=$1
+  sanitize_inputfile file_to_read
```
goranche: replace `file_to_read` with `${file_to_read}`  :wink:

goranche: and maybe quotes? :thought_balloon:

scripts/etc/init.d/rcS
```diff
@@ -85,6 +85,25 @@ sanitize_inputfile()
...
+  grep -v "^[[:space:]]*#\|^[[:space:]]*$" "${file_to_read}" | while read -r line; do
+    owner=$(echo "$line" | awk '{ print $1 }')
+    perms=$(echo "$line" | awk '{ print $2 }')
+    file=$(echo  "$line" | awk '{ print $3 }')
+    if [ ! -d "config/files$file" ]; then
+      mkdir -p "/rootfs$(dirname "$file")"
+      cp "config/files$file" "/rootfs$file"
+    else
+      mkdir -p "../rootfs/$file"
```
goranche: it seems you really like the double dots :wink: 
but I don't think `..` should be here

Mausy5043: AAAARGH!!!!

scripts/etc/init.d/rcS
Mausy5043: good catch

Mausy5043: Why `{  }`?

goranche: readabilit, and it's "safer"... (we should probably change this throughout the file, but ok)

for example, what if you want to do this: `"${foo}bar"`
if you want to use parameter expansion operations, you have to use the curly braces (`${filename%.*}` - removes the extension)
...
what you start using arrays (ok, ash doesn't support them, I know), you have to use the curlies...

I just use them always, makes things more readable, but here they aren't really needed

diederikdehaas: > we should probably change this throughout the file

:+1:

Mausy5043: Well, there's no danger here, we're not using arrays and the rest of the file doesn't use curly braces either. I want to keep it like this.
If you want to add curly braces we should do all scripts and that is/should be IMO a separate issue.

goranche: > that is/should be IMO a separate issue

agree, it will be... :wink:   (that's why I thought it wouldn't hurt if you change it now)

Mausy5043: I rather not in this PR.

scripts/etc/init.d/rcS
```diff
@@ -85,6 +85,25 @@ sanitize_inputfile()
     fi
 }
 
+install_files()
+{
+  file_to_read=$1
```
goranche: could you add double quotes around `$1`? :innocent:

goranche:
I know you'll probably hate me, but... :innocent: 

It would be nice to have a mention that you're supposed to call the install_files function...

or, we could also just provide a flag to the installer... and when set to true (or 1, which is our convention, I think), it would call install_files automagically... :thought_balloon: 
just before running post-install.txt, for example

Mausy5043:
I'll start thinking about some text for the README.md and add that to the commit.
It'll be something along the lines of end-user having to call the function with the appropriate filename from `post-install.txt`.

README.md
```diff
@@ -174,6 +174,16 @@ Presets set the `cdebootstrap_cmdline` variable. For example, the current _serve
...
+You can have the installer place your custom configuration files (or any other file you wish to add) on the installed system during the installation. For this, you need to provide the necessary files in the `/config/files/` directory of your SD card (you may need to create this directory if it doesn't exist). The `/config/files/` directory is the root-point. It must have the same structure as inside the installed system. So, a file that you place on the SD card in `/config/files/etc/wpa_supplicant/wpa_supplicant.conf` will end up on the installed system as `/etc/wpa_supplicant/wpa_supplicant.conf`.
+Each file or directory that you wish to place on the target system must also be listed in a configuration file in the directory `/config` on your SD card. This allows you to specify the owner (and group) and the permissions of the file. An example file is provided with the installer (see `/config/my-files.list` for more information). ONLY files listed there are copied over to the installed system.
+To have the installer actually copy the files to the target system, add the following command at an appropriate point in your `post-install.txt` file:
+```
+install_files my-files.list
```
goranche: change this to include a path to the file... and maybe change the filename to `<my-files.list>` to indicate that it can be any filename :thought_balloon:

goranche: on the other hand, my-files.list is provided as an example... meh, let it be, sorry :smile:

README.md
```diff
@@ -174,6 +174,16 @@ Presets set the `cdebootstrap_cmdline` variable. For example, the current _serve
...
+```
+  where `my-files.list` is the name of the file containing the list of files.
+Please be aware that some restrictions may apply to the sum of the file sizes. If you wish to supply large files in this manner you may need to adjust the value of the `bootsize` parameter.
+
```
goranche: maybe add a paragraph explaining that the function can be called multiple times if needed, with different list files...

Mausy5043:
Okay, it's working using absolute paths.

scripts/etc/init.d/rcS
```diff
@@ -85,6 +85,28 @@ sanitize_inputfile()
...
+    if [ ! -d "/bootfs/config/files$file" ]; then
+      mkdir -p "/rootfs$(dirname "$file")"
+      cp "/bootfs/config/files$file" "/rootfs$file"
+    else
+      mkdir -p "/rootfs/$file"
+    fi
+    chmod "$perms" "/rootfs$file"
+    chroot /rootfs chown "$owner" "$file"
+  done
+  echo " "
```
goranche: what is this single space echo for?

scripts/etc/init.d/rcS
```diff
@@ -85,6 +85,28 @@ sanitize_inputfile()
...
+install_files()
+{
+  file_to_read="$1"
+  echo "Adding files & folders listed in /bootfs/config/$file_to_read"
+  sanitize_inputfile "/bootfs/config/$file_to_read"
+  grep -v "^[[:space:]]*#\|^[[:space:]]*$" "/bootfs/config/${file_to_read}" | while read -r line; do
+    owner=$(echo "$line" | awk '{ print $1 }')
+    perms=$(echo "$line" | awk '{ print $2 }')
+    file=$(echo  "$line" | awk '{ print $3 }')
+    echo "$file"
```
goranche: could you add two or four spaces in front of `$file'? that way the list will be indented in the log file and easier to scan through

scripts/etc/init.d/rcS
Mausy5043: empty line

goranche:
@Mausy5043 thanks, great work :+1: 
looking forward to changing my `post-install.txt` scripts to make use of this

Mausy5043:
This PR should can help fix issue #349

goranche:
yes, this PR is a great step forward

PR #396, opened 2016-03-27
wpa_supplicant correction

Mausy5043:
Attempting to fix issue #349 
Moving the required location of `wpa_supplicant.conf` into the `/bootfs/config/files/...` tree.
This enables the file to be installed automagically on the target system.
At the same time the file can be used by the installer to support WiFi installation.


Mausy5043:
Also related to issue #44

goranche:
but just placing the file in `config/files/...` will not "automagically" copy the file to the installed system...

Mausy5043:
I know you have to add it to the file list from #394

Mausy5043:
A different approach is required. Closing.

Mausy5043:
Not yet deleting the branch

PR #397, opened 2016-03-28
wpa_supplicant usage

Mausy5043:
Changes:
  1. Put all networking related code together.
  2. If `config/wpa_supplicant.conf` exists AND was used by the installer then copy it to the installed system in `/etc/wpa_supplicant/` and add a reference to it in `/etc/network/interfaces`.
  3. Changed the way of installing and starting `wpa_supplicant` so that it actually works.

This commit will allow the user to **install** AND **use** the target system in the absence of a wired ethernet connection, but in the presence of a WiFi network.

The user must configure the following (already existing) parameters:
  1. `drivers_to_load=` supply the name of the module that must be loaded by the kernel for the hardware (WiFi dongle) used. _E.g.: `8192cu`_.
  1. `ifname=` supply the non-ethernet interface name. Normally you will use `wlan0` here.  

Additionally user must supply a file (`config/wpa_supplicant.conf`) containing a valid configuration. E.g.:  
> network={
>        ssid="pi3.1415926"
>        psk="AV3ry53cretPa55w0rdThatN00n3W!113verGue55"
>        key_mgmt=WPA-PSK
> }

ref: https://w1.fi/cgit/hostap/plain/wpa_supplicant/wpa_supplicant.conf
(Please don't use the password suggested in the example above. It is not secure)

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -1117,6 +1117,15 @@ if [ -e /bootfs/post-install.txt ]; then
     echo "================================================="
 fi
 
+# modify installed network settings
+if [ -f /rootfs/etc/wpa_supplicant/wpa_supplicant.conf ]; then
+    if [ $(grep "iface" /rootfs/etc/network/interfaces |grep -v "lo\|eth0" |wc -l) -ne 0 ]; then
+        if [ $(grep "wpa[-_]" /rootfs/etc/network/interfaces |wc -l) -eq 0 ]; then
+            echo "wpa-conf /etc/wpa_supplicant/wpa_supplicant.conf" >> /rootfs/etc/network/interfaces
+        fi
+    fi
+fi
+
 # remove cdebootstrap-helper-rc.d which prevents rc.d scripts from running
 echo -n "Removing cdebootstrap-helper-rc.d... "
 chroot /rootfs /usr/bin/dpkg -r cdebootstrap-helper-rc.d &>/dev/null || fail
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -838,29 +838,43 @@ echo "OK"
 # networking
 echo -n "  Configuring network settings... "
 touch /rootfs/etc/network/interfaces || fail
-{
-    # wheezy seems to add the lo interface, so first check for it
-    if ! grep -q "auto lo" /rootfs/etc/network/interfaces; then
-        echo "auto lo"
-        echo "iface lo inet loopback"
-        echo ""
-    fi
+# lo interface may already be there, so first check for it
+if ! grep -q "auto lo" /rootfs/etc/network/interfaces; then
+    echo "auto lo"                      >> /rootfs/etc/network/interfaces
+    echo "iface lo inet loopback"       >> /rootfs/etc/network/interfaces
+    echo ""                             >> /rootfs/etc/network/interfaces
+fi
 
-    # eth0 config
-    echo "auto $ifname"
-    if [ "$ip_addr" = "dhcp" ]; then
-        echo "iface $ifname inet dhcp"
+# eth0 config
+echo "auto $ifname"                     >> /rootfs/etc/network/interfaces
+if [ "$ip_addr" = "dhcp" ]; then
+    echo "iface $ifname inet dhcp"      >> /rootfs/etc/network/interfaces
+else
+    echo "iface $ifname inet static"    >> /rootfs/etc/network/interfaces
+    echo "    address $ip_addr"         >> /rootfs/etc/network/interfaces
+    echo "    netmask $ip_netmask"      >> /rootfs/etc/network/interfaces
+    echo "    broadcast $ip_broadcast"  >> /rootfs/etc/network/interfaces
+    echo "    gateway $ip_gateway"      >> /rootfs/etc/network/interfaces
+fi
+
+# wlan config
+if [ "$ifname" != "eth0" ]; then
+    # Install wireless dependencies
+    echo -n "Installing wpasupplicant package... "
+    chroot /rootfs /usr/bin/apt-get -y install wpasupplicant >& /dev/null
+    if [ $? -eq 0 ]; then
+        echo "OK"
     else
-        echo "iface $ifname inet static"
-        echo "    address $ip_addr"
-        echo "    netmask $ip_netmask"
-        echo "    broadcast $ip_broadcast"
-        echo "    gateway $ip_gateway"
+        echo "FAILED !"
     fi
-    if [ "$ifname" != "eth0" ]; then
-        echo "wpa-conf /boot/config/wpa_supplicant.conf"
+    # 
+    if [ -e /bootfs/config/wpa_supplicant.conf ]; then
+        # copy the installer version of `wpa_supplicant.conf`
+        cp /bootfs/config/wpa_supplicant.conf /rootfs/etc/wpa_supplicant/
     fi
-} >> /rootfs/etc/network/interfaces || fail
+    echo "wpa-conf /boot/config/wpa_supplicant.conf"	>> /rootfs/etc/network/interfaces
+fi
+
 
 if [ "${disable_predictable_nin}" = "1" ]; then
     # as described here: https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames
@@ -1024,18 +1038,6 @@ if [ $? -eq 0 ]; then
     echo "OK"
 else
     echo "FAILED !"
-fi
-
-# if eth0 was not used during initial boot, a wireless interface was used
-if [ "$ifname" != "eth0" ]; then
-    # Install wireless dependencies
-    echo -n "Installing wpasupplicant package... "
-    chroot /rootfs /usr/bin/apt-get -y install wpasupplicant >& /dev/null
-    if [ $? -eq 0 ]; then
-        echo "OK"
-    else
-        echo "FAILED !"
-    fi
 fi
 
 if [ "$hwrng_support" = "1" ]; then
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -866,6 +866,7 @@ fi
 if [ "$ifname" != "eth0" ]; then
     # Install wireless dependencies
     echo -n "Installing wpasupplicant package... "
+    mkdir -p /rootfs/etc/wpa_supplicant/
     chroot /rootfs /usr/bin/apt-get -y install wpasupplicant >& /dev/null
     if [ $? -eq 0 ]; then
         echo "OK"
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -842,14 +842,15 @@ touch /rootfs/etc/network/interfaces || fail
 if ! grep -q "auto lo" /rootfs/etc/network/interfaces; then
     echo "auto lo" >> /rootfs/etc/network/interfaces
     echo "iface lo inet loopback" >> /rootfs/etc/network/interfaces
-    echo  >> /rootfs/etc/network/interfaces
 fi
 
 # eth0 config
 echo "auto $ifname" >> /rootfs/etc/network/interfaces
 if [ "$ip_addr" = "dhcp" ]; then
+    echo  >> /rootfs/etc/network/interfaces
     echo "iface $ifname inet dhcp" >> /rootfs/etc/network/interfaces
 else
+    echo  >> /rootfs/etc/network/interfaces
     echo "iface $ifname inet static" >> /rootfs/etc/network/interfaces
     echo "    address $ip_addr" >> /rootfs/etc/network/interfaces
     echo "    netmask $ip_netmask" >> /rootfs/etc/network/interfaces
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -842,7 +842,7 @@ touch /rootfs/etc/network/interfaces || fail
 if ! grep -q "auto lo" /rootfs/etc/network/interfaces; then
     echo "auto lo" >> /rootfs/etc/network/interfaces
     echo "iface lo inet loopback" >> /rootfs/etc/network/interfaces
-    echo ""  >> /rootfs/etc/network/interfaces
+    echo  >> /rootfs/etc/network/interfaces
 fi
 
 # eth0 config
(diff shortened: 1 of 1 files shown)


scripts/etc/init.d/rcS
```diff
@@ -838,29 +838,43 @@ echo "OK"
...
-{
-    # wheezy seems to add the lo interface, so first check for it
-    if ! grep -q "auto lo" /rootfs/etc/network/interfaces; then
-        echo "auto lo"
-        echo "iface lo inet loopback"
-        echo ""
-    fi
+# lo interface may already be there, so first check for it
+if ! grep -q "auto lo" /rootfs/etc/network/interfaces; then
+    echo "auto lo"                      >> /rootfs/etc/network/interfaces
```
goranche: please don't use weird alignment like this

scripts/etc/init.d/rcS
```diff
@@ -838,29 +838,43 @@ echo "OK"
...
+    echo "iface lo inet loopback"       >> /rootfs/etc/network/interfaces
+    echo ""                             >> /rootfs/etc/network/interfaces
```
goranche: when echo-ing an empty line, no quotes are needed (or desired, as it makes it clear(er) that the intention was an empty line)...

that said, I don't think an empty line is needed here at all... if anywhere, add it before adding additional "entries"

scripts/etc/init.d/rcS
```diff
@@ -838,29 +838,43 @@ echo "OK"
...
-        echo "iface $ifname inet static"
-        echo "    address $ip_addr"
-        echo "    netmask $ip_netmask"
-        echo "    broadcast $ip_broadcast"
-        echo "    gateway $ip_gateway"
+        echo "FAILED !"
     fi
-    if [ "$ifname" != "eth0" ]; then
-        echo "wpa-conf /boot/config/wpa_supplicant.conf"
+    #
```
goranche: remove this empty comment line

scripts/etc/init.d/rcS
```diff
@@ -838,29 +838,43 @@ echo "OK"
...
     fi
-    if [ "$ifname" != "eth0" ]; then
-        echo "wpa-conf /boot/config/wpa_supplicant.conf"
+    # 
+    if [ -e /bootfs/config/wpa_supplicant.conf ]; then
+        # copy the installer version of `wpa_supplicant.conf`
+        cp /bootfs/config/wpa_supplicant.conf /rootfs/etc/wpa_supplicant/
     fi
-} >> /rootfs/etc/network/interfaces || fail
+    echo "wpa-conf /etc/wpa_supplicant/wpa_supplicant.conf"	>> /rootfs/etc/network/interfaces
```
goranche: remove this line, and after running `post-install.txt`, check for the existence of the `/rootfs/etc/wpa_supplicant/wpa_supplicant.conf` file (it might be put there using `install_files`, even if not used in the installer as such)... if found, add the appropriate entry to `/rootfs/etc/network/interfaces`

Mausy5043: Not agreeing with this. The "wpa-conf" must be added if the .conf file is added. So moving this line inside the `if..fi` instead.

IMO, if the end-user adds a `wpa_supplicant.conf` using `install_files` they should also check (and adjust if necessary) the `/etc/network/interfaces` file via the `post-install.txt`.

goranche: > The "wpa-conf" must be added if the .conf file is added

it **will** be added by the code after `post-install.txt`... the difference is just that it needs to be done only once, without putting the burden of checking the interfaces file on the end user

goranche: come to think of it... we should check if the user has added anything to the interfaces file :thought_balloon:

Mausy5043:
bump

Mausy5043:
This PR fixes #175 which IMHO was never properly tested (due to lack of hardware for testing).
This PR solves #44 and #349.
This PR may be of help finding a solution for #52

scripts/etc/init.d/rcS
```diff
@@ -272,9 +272,13 @@ if [ "$ifname" != "eth0" ]; then
     # Replace eth0 as udhcpc dns interface
     sed -i "s/PEERDNS_IF=.*/PEERDNS_IF=$ifname/g" /etc/udhcpc/default.script
     # if eth0 is not the chosen eth interface it is a wireless interface and wpa_supplicant must connect to wlan
-    echo -n "Running wpa_supplicant... "
-    wpa_supplicant -B -Dnl80211,wext -c/bootfs/config/wpa_supplicant.conf -i$ifname >& /dev/null || fail
-    echo "OK"
+    echo "Starting wpa_supplicant... "
+    wpa_supplicant -B -Dnl80211 -c/bootfs/config/wpa_supplicant.conf -i$ifname
```
goranche: are `nl80211` and `wext` the only two supported modules?

scripts/etc/init.d/rcS
```diff
@@ -272,9 +272,13 @@ if [ "$ifname" != "eth0" ]; then
...
+    if [ $? -ne 0 ]; then
+        echo "nl80211 driver is not working. Trying generic driver (wext)..."
```
goranche: "didn't work" instead of "is not working"... 😉

scripts/etc/init.d/rcS
```diff
@@ -409,6 +413,15 @@ case "$rootfstype" in
     ;;
 esac
 
+# check if we need wpasupplicant package
```
goranche: the comment is a bit misleading, as this isn't really a check if the packages is needed, but if an interface other than `eth0` is used 😉

scripts/etc/init.d/rcS
```diff
@@ -838,29 +851,34 @@ echo "OK"
...
-        echo "    netmask $ip_netmask"
-        echo "    broadcast $ip_broadcast"
-        echo "    gateway $ip_gateway"
-    fi
-    if [ "$ifname" != "eth0" ]; then
-        echo "wpa-conf /boot/config/wpa_supplicant.conf"
+# eth0 config
+echo "auto $ifname" >> /rootfs/etc/network/interfaces
+if [ "$ip_addr" = "dhcp" ]; then
+    echo  >> /rootfs/etc/network/interfaces
```
goranche: why this empty line into `interfaces`?

scripts/etc/init.d/rcS
Mausy5043: There's also `wired`.

scripts/etc/init.d/rcS
```diff
@@ -838,29 +851,34 @@ echo "OK"
...
+    echo "    broadcast $ip_broadcast" >> /rootfs/etc/network/interfaces
+    echo "    gateway $ip_gateway" >> /rootfs/etc/network/interfaces
+fi
+
+# wlan config
+if [ "$ifname" != "eth0" ]; then
+    if [ -e /bootfs/config/wpa_supplicant.conf ]; then
+        # copy the installer version of `wpa_supplicant.conf`
+        cp /bootfs/config/wpa_supplicant.conf /rootfs/etc/wpa_supplicant/
+        echo "wpa-conf /etc/wpa_supplicant/wpa_supplicant.conf"	>> /rootfs/etc/network/interfaces
```
goranche: shouldn't the `wpa-conf` option come after a `iface` definition line? 💭 

this is what I have in my `interfaces` file (taken from the latest raspbian):
```
allow-hotplug wlan0
iface wlan0 inet manual
    wpa-conf /etc/wpa_supplicant/wpa_supplicant.conf
```

scripts/etc/init.d/rcS
Mausy5043: But we don't need `wpa_supplicant` for wired connections :wink:

scripts/etc/init.d/rcS
```diff
@@ -1115,6 +1121,15 @@ if [ -e /bootfs/post-install.txt ]; then
     echo "================================================="
 fi
 
+# modify installed network settings
+if [ -f /rootfs/etc/wpa_supplicant/wpa_supplicant.conf ]; then
+    if [ $(grep "iface" /rootfs/etc/network/interfaces |grep -v "lo\|eth0" |wc -l) -ne 0 ]; then
```
goranche: please add spaces after the pipe characters (the actual pipes, not the character in the regex 😉 )

scripts/etc/init.d/rcS
Mausy5043: Good catch

scripts/etc/init.d/rcS
Mausy5043: Changed

scripts/etc/init.d/rcS
Mausy5043: And we assume that this is a wireless interface and therefore the `wpasupplicant` package is needed.

goranche:
💭 maybe it would be better to check the interface name for `wlan*` instead of `not eth0`...
a possible problem I see is it anyone has additional non-wifi interfaces...

scripts/etc/init.d/rcS
Mausy5043: It does.
If `ip_addr` != "dhcp" then there a a few lines inbetween. That's not a problem.

scripts/etc/init.d/rcS
goranche: 😋  I meant for wireless... 😊

diederikdehaas:
> better to check the interface name for `wlan*` instead of not `eth0`

Ra-link devices get named `ra*` and those are wireless devices. 
What do you mean with _additional non-wifi interfaces_ (that could pose a problem)?

scripts/etc/init.d/rcS
goranche: I know... I added a comment on this in the "general" discussion part...

Mausy5043:
> a possible problem I see is it anyone has additional non-wifi interfaces...

Let's cross that bridge when we get there.

scripts/etc/init.d/rcS
diederikdehaas: What's the reason for this change?
The whole idea of `-Dnl80211,wext` is to try `nl80211` first and if that doesn't work, try `wext`.

scripts/etc/init.d/rcS
goranche: it's just "ugly" 😁   maybe we should at least add indentation, so it's a bit more obvious what belongs together

scripts/etc/init.d/rcS
Mausy5043: It doesn't work with `-Dnl80211,wext` 😕 . It turned out during testing that it only worked if I tried them separately.

goranche:
> What do you mean with additional non-wifi interfaces (that could pose a problem)?

I was thinking a bit too far ahead... this is a larger part of my `interfaces file`:
```
allow-hotplug wlan0
iface wlan0 inet manual
    wpa-conf /etc/wpa_supplicant/wpa_supplicant.conf

allow-hotplug wlan1
iface wlan1 inet manual
    wpa-conf /etc/wpa_supplicant/wpa_supplicant.conf
```

the code as is wouldn't create both wpa-conf entries... but on the other hand (this is where I didn't think it through) you can't really add multiple interfaces anyway, so... never mind 😉

goranche:
> Ra-link devices get named ra*

*groan*

scripts/etc/init.d/rcS
```diff
@@ -838,29 +851,33 @@ echo "OK"
...
-        echo "    address $ip_addr"
-        echo "    netmask $ip_netmask"
-        echo "    broadcast $ip_broadcast"
-        echo "    gateway $ip_gateway"
-    fi
-    if [ "$ifname" != "eth0" ]; then
-        echo "wpa-conf /boot/config/wpa_supplicant.conf"
+# eth0 config
+echo "" >> /rootfs/etc/network/interfaces
+echo "auto $ifname" >> /rootfs/etc/network/interfaces
```
diederikdehaas: With wireless interfaces, you usually also want `allow-hotplug $ifname` as wireless devices are more likely to get unplugged and replugged, but also it will give a signal when the device is ready to be used by which the `auto` may already have passed.

goranche:
I do have one more thing, though... if we install using WiFi, we don't create an entry for eth0...

I think that in this case we should create an entry with dhcp:
```
allow-hotplug eth0
iface eth0 inet dhcp
```

this way the ethernet port will still work if a cable is plugged in...

diederikdehaas:
I'm not sure where to put it, so I'll just do it here.
The wireless part of my `/e/n/i` is as follows:
```
allow-hotplug wlan0
iface wlan0 inet manual
wpa-roam /etc/wpa_supplicant/wpa_supplicant.conf
iface default inet dhcp
```
Not the `manual` value and the extra `iface ...` line at the bottom

scripts/etc/init.d/rcS
Mausy5043: aah, yeah you're right

goranche:
> Not the manual value and the extra iface ... line at the bottom

I have manual in there as well (interestingly, for eth0 as well, as said, taken from the official raspbian)...

what is a `default` iface?

diederikdehaas:
> what is a default iface?

I actually don't know, I just know that I see that all the time and it also always works here.

Mausy5043:
>> what is a default iface?
> I actually don't know, I just know that I see that all the time and it also always works here.

I've never used it. So, I hope you don't mind me not adding it. It works fine without.
Agree with the `auto` and `hotplug` entries. Will add those.

Mausy5043:
Mind you that not adding the `eth0` interface here, doesn't mean it won't work.
It actually still will.
So, I'm adding entries for `eth0` per your (2) request. But, it is not needed.

Mausy5043:
Done.

goranche:
> not adding the eth0 interface here, doesn't mean it won't work

huh? really? why? how? I NEED ANSWERS!!! 😁 

no, but really... how?

Mausy5043:
I have no idea. I just does. Just guessing, but could it be `dhclient`?

goranche:
> could it be dhclient

it shouldn't be, AFAIK, if the interface isn't listed (to be brought up) in the `interfaces` file... 💭

diederikdehaas:
FWIW I share goranche's skepticism.

Mausy5043:
OK. No problem.  The `eth0` definition has already been added as you requested so ... the discussion is moot anyway,

diederikdehaas:
Next time I will require that you'll squash commits together :wink: 

I noticed that when you only have a wireless device plugged in, the boot process takes quite a long time because it's waiting for `eth0`, so you (@Mausy5043) may have had more of a point then I anticipated.

PR #408, opened 2016-06-16
Kernel 4.4 update

diederikdehaas:
Currently you can't build an installer that boots due to incompatibility between the new firmware files and the 3.18 kernel. This PR fixes that.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -5,12 +5,8 @@ set -e
 KERNEL_VERSION_RPI1=4.4.0-1-rpi
 KERNEL_VERSION_RPI2=4.4.0-1-rpi2
 
-INSTALL_MODULES="kernel/fs/f2fs/f2fs.ko"
-INSTALL_MODULES="$INSTALL_MODULES kernel/fs/btrfs/btrfs.ko"
-INSTALL_MODULES="$INSTALL_MODULES kernel/drivers/char/hw_random/bcm2708-rng.ko"
-INSTALL_MODULES="$INSTALL_MODULES kernel/drivers/usb/storage/usb-storage.ko"
+INSTALL_MODULES="kernel/fs/btrfs/btrfs.ko"
 INSTALL_MODULES="$INSTALL_MODULES kernel/drivers/scsi/sg.ko"
-INSTALL_MODULES="$INSTALL_MODULES kernel/drivers/scsi/sd_mod.ko"
 
 # checks if first parameter is contained in the array passed as the second parameter
 #   use: contains_element "search_for" "${some_array[@]}" || do_if_not_found
--- a/build.sh
+++ b/build.sh
@@ -2,8 +2,8 @@
 
 set -e
 
-KERNEL_VERSION_RPI1=3.18.0-trunk-rpi
-KERNEL_VERSION_RPI2=3.18.0-trunk-rpi2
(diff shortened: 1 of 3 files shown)


Mausy5043:
This change will prevent use of pre-4.x kernels. Is that intended?

diederikdehaas:
Can you clarify what you mean (which change)?

A change/update in the `raspberrypi-bootloader-nokernel` package caused the 3.18 kernel to not boot anymore. So without some workaround to use an older version of that package, you can't use kernel 3.18 anymore. I haven't tested older kernels then 3.18 though.

So it's changes in the archives that may indeed prevent the use of older kernels.
This change just adapts adapts to that. I would have preferred that any old kernel would still be usable, but I fear that it is not possible anymore.

Mausy5043:
I meant ripping out the modules that are built-in into the kernel 4.x

Mausy5043:
I did an install with the current v1.1.x installer this morning and ended up with a system that has:
```
pi@rbian ~ $ uname -a
Linux rbian 4.4.0-1-rpi #1 Debian 4.4.6-1+rpi14 (2016-05-05) armv6l GNU/Linux
```

So, what's up with that then?

Mausy5043:
Oh wait. The installed system gets 4.4.0.  The installer not yet. I get it now

diederikdehaas:
Then yes. Or we make it all conditionally which is probably a bad idea as it's a lot of work for very little gain. At least for the installer kernel, which this PR is about.
But yes, the installed system gets the 4.4 kernel too as the kernel meta package was updated too.
Plugwash didn't intended to do that yet, but did so when I reported him that the 3.18 kernel didn't boot anymore.

I think it is a terrible change by the RPF, which raspbian.org 'inherited'. For some stupid reason they still refuse to use an initramfs and thus change the kernel config at random, causing issues like you see in this PR.

Mausy5043:
Well, my original question was just to make sure that this change is a deliberate move. I don't see a need to question the change itself.

diederikdehaas:
I am not happy that the change is needed, but it is needed for the 4.4 kernel and therefor my code changes were deliberate.

Mausy5043:
:ok:
Not that it matters a lot, but any reason you opted for the 4.4.0 i.s.o. a newer build?

diederikdehaas:
There's a big difference in the version numbering scheme the RPF uses and what Debian/raspbian.org uses. 
The RPF actually doesn't have one apart from \<previous-nr\> + 1, but Debian/Raspbian versions their kernels according to stable ABI. 

Earlier you posted:
`Linux rbian 4.4.0-1-rpi #1 Debian 4.4.6-1+rpi14 (2016-05-05) armv6l GNU/Linux`
Which means it's actually using the 4.4.6 kernel version.

diederikdehaas:
Also: I'm not picking a specific kernel (version) myself, I just use what's made available in the raspbian.org repos.

Mausy5043:
Great job @diederikdehaas. Tanks 👍 .

Mausy5043:
Migrating to newer kernels (which is a good thing) makes it more difficult to downgrade to older kernels. Should we maybe close #166 ?

diederikdehaas:
I agree that you'd likely *want* newer kernels, but I still consider #166 a valid request and it shouldn't actually be too difficult to implement as raspbian.org still has those older kernels in the repo.

PR #413, opened 2016-06-30
add license title

waldyrious:
It's not strictly required, but it's useful metadata, and part of the recommended license template text (see http://choosealicense.com/licenses/isc/ and https://opensource.org/licenses/isc-license)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,5 @@
+ISC License
+
 Copyright (c) 2013 Toni Spets <toni.spets@iki.fi>
 
 Permission to use, copy, modify, and distribute this software for any
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,6 @@
-ISC License
+# SPDX-License-Identifier: ISC
+
+ISC License:
 
 Copyright (c) 2013 Toni Spets <toni.spets@iki.fi>
 


diederikdehaas:
If you make it a [SPDX License](https://spdx.org/licenses/) identifier (and squash the 2 commits into 1), we'll have a deal :smile:

waldyrious:
Sure, but note that the [SPDX sample text](https://spdx.org/licenses/ISC.html#licenseText) actually uses "ISC License" (with a colon in the end, which can be added here too). So wouldn't that be the appropriate choice? c.f. [GPL](http://spdx.org/licenses/GPL-3.0.html#licenseText), [Apache](http://spdx.org/licenses/Apache-2.0.html#licenseText), [Artistic](http://spdx.org/licenses/Artistic-2.0.html#licenseText).

diederikdehaas:
From what I've understood of it, you're supposed to use the (short) ID(entifier), not the name.
See https://github.com/david-a-wheeler/spdx-tutorial/#spdx-license-expressions-in-source-code-files, http://wiki.spdx.org/view/Technical_Team/SPDX_Meta_Tags and https://events.linuxfoundation.org/sites/events/files/slides/DevelopersCare_JLovejoy_LinuxConEu_2015_final-static.pdf

diederikdehaas:
Let's do both.
I was already thinking about adding the SPDX-Licence-ID to it, but your point is correct as well.
So start with a comment block with the ID, like
```
# SPDX-License-Identifier: ISC
``` 
followed by `ISC License:` and that followed by the (original) text.

waldyrious:
sounds good :)

waldyrious:
@diederikdehaas done. If you prefer them squashed, note that this can now be done [directly from the github interface](https://github.com/blog/2141-squash-your-commits) (but I don't think it's a bad idea to have both commits as each one is indeed semantically valid as a standalone change).

waldyrious:
(by the way, thanks for the links -- I've bookmarked them)

diederikdehaas:
Thanks :+1: 

There's no need for you to squash the commit. I do think they belong into one commit though.
But I'm currently writing a Git(Hub) workflow guide for someone and this is a perfect practical case to apply it too :wink: 
Merging of your PR may therefor take some time as the guide isn't finished yet.

waldyrious:
Ok then :) merge whenever it's appropriate. Cheers!

diederikdehaas:
Not sure I'll continue/finish that guide, so better merge this PR.
Thanks 👍

waldyrious:
My pleasure :)

PR #414, opened 2016-07-14
Add dosfstools to base_packages

Phiber2000:
Package `dosfstools` should be added to base_packages because mmcblk0p1 is FAT formatted.
`util-linux`, which is included in `linux-base` and includes `fsck` compatibility for `ext4`. So I think `dosfstools` is placed correctly in base_packages.

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -442,7 +442,7 @@ if [ "$cdebootstrap_cmdline" = "" ]; then
     fi
 
     # base
-    base_packages="cpufrequtils,kmod"
+    base_packages="cpufrequtils,kmod,dosfstools"
     for hv in ${hardware_versions}
     do
         case $hv in


goranche:
thanks for your contribution 👍 

I'm not sure I agree with this one, although on the other hand, the fsck might come in handy, I agree here... 💭 
will take a look 😅

diederikdehaas:
The `dosfstools` package is already part of the `minimal_packages` variable and I tend to think that is the proper place. The `base_packages` variable is more for packages without which your system wouldn't work properly.

goranche:
well, good, I don't need to check anything... 😉

diederikdehaas:
You still can. You can even disagree with me.

Damn, I must be in a good mood today :smile:

goranche:
disagree regarding what? 😄   I'd actually remove dosfstools from the installation altogether... 😁 

although, it's been in there till now, not sure it makes sense to remove it 😉

PR #424, opened 2016-09-05
Remove note about installing rng-tools as hwrng_support is on by default

kpfleming:
If the user has not overridden it, hwrng_support is on by default, and the
rng-tools package will be installed. In addition, in current kernels the
hwrng driver is in the kernel, not delivered as a module. If the user *has*
overridden the hwrng_support default, they don't need a note about how to
install rng-tools because they presumably did that on purpose :-)


Mausy5043:
@kpfleming Thanks!

PR #425, opened 2016-09-10
Add IPv6 support

kpfleming:
This patch adds substantial support for use of IPv6 in the installer,
and in the installed system. It supports static configuration, although
most IPv6 networks use automatic (and dynamic) configuration. This
patch does not provide DHCPv6 support as there is no suitable DHCPv6
client available to use in the installer's environment, but it does
provide support for Router Advertisements, SLAAC, and RDNSS.

In order to simplify management of DNS resolver information in the
installed system (since there are multiple potential sources of this
information), this patch adds the 'resolvconf' package to the installed
system. If IPv6 DNS information is gathered from RDNSS, the 'rdnssd'
package will also be added to the installed system.

For discovery of Router Advertisment-provided information during the
installation, the patch also adds tools from the 'ndisc6' package to
the installer.

Finally, to provide consistent syntax in the installer configuration
file, this patch changes the 'ip_*' options to use an 'ip4_' prefix
instead, and also uses 'prefix length' instead of 'netmask' for
IPv4 networks. It supports prior configuration files, though, and
will automatically convert options in the old format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,9 @@ For those unfamiliar with IPv6 networking, there are some significant difference
 IPv6 networks can provide automatic addressing, automatic gateway discovery, and automatic DNS resolver discovery, but these can be provided
 independently. As a result, configuration of the installer for IPv6 is done in three parts.
 
+The simplest configuration is when the network supports SLAAC and RDNSS; this is roughly equivalent to IPv4 DHCP, and the installer will be
+able to automatically assign an address, gateway, and get DNS resolver address(es).
+
 #### Addressing
 
 The installer can be configured in three modes:
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ The default for IPv4 is to use DHCP to obtain an address/prefix, default gateway
 ### IPv6
 
 The default for IPv6 is to disable its use; even if the network advertises IPv6 information, it will not be used. Note that DHCPv6 is *not*
-supported, as there is no suitable DHCPv6 client available for us in the installer environment. If the network indicates that DHCPv6 is required
+supported, as there is no suitable DHCPv6 client available for use in the installer environment. If the network indicates that DHCPv6 is required
 for addressing or any other network information, the installer will not use IPv6. The installed system can use DHCPv6, but the installer
 is unable to configure it in that mode.
 
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ modes:
   do not provide DNS resolution over IPv6.
 
 ## Logging
-The output of the installation process is now also logged to file.
+The output of the installation process is now also logged to file.  
 When the installation completes successfully, the logfile is moved to /var/log/raspbian-ua-netinst.log on the installed system.  
 When an error occurs during install, the logfile is moved to the sd card, which gets normally mounted on /boot/ and will be named raspbian-ua-netinst-\<datetimestamp\>.log
 
@@ -300,8 +300,8 @@ The system is almost completely unconfigured on first boot. Here are some tasks 
 The default **root** password is **raspbian**.
 
 > Set new root password: `passwd`  (can also be set during installation using **rootpw** in [installer-config.txt](#installer-customization))  
-> Configure your default locale: `dpkg-reconfigure locales`
-> Configure your timezone: `dpkg-reconfigure tzdata`
+> Configure your default locale: `dpkg-reconfigure locales`  
+> Configure your timezone: `dpkg-reconfigure tzdata`  
 
 The latest kernel and firmware packages are now automatically installed during the unattended installation process.
 When you need a kernel module that isn't loaded by default, you will still have to configure that manually.
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 - [Writing the installer to the SD card](#writing-the-installer-to-the-sd-card)
 - [Installing](#installing)
 - [Installer customization](#installer-customization)
+- [IP Networking](#ip-networking)
 - [Logging](#logging)
 - [First boot](#first-boot)
 - [Reinstalling or replacing an existing system](#reinstalling-or-replacing-an-existing-system)
@@ -141,11 +142,14 @@ The format of the _installer-config.txt_ file and the current defaults:
                               # RPi board, your network device might be named differently. This will result in the
                               # board having no network connectivity.
     ifname=eth0
-    ip_addr=dhcp
-    ip_netmask=0.0.0.0
-    ip_broadcast=0.0.0.0
-    ip_gateway=0.0.0.0
-    ip_nameservers=
+    ip4_addr=dhcp # options are 'disable', 'dhcp', or an IPv4 address
+    ip4_prefixlength=0
+    ip4_gateway=0.0.0.0
+    ip4_nameservers=
+    ip6_addr=disable # options are 'disable', 'auto', or an IPv6 address
+    ip6_prefixlength=0
+    ip6_gateway=auto # options are 'auto', or an IPv6 address (which will only be applied if ip6_addr is a static address)
+    ip6_nameservers=auto # options are 'auto', 'disable', or an IPv6 address
     drivers_to_load=
     online_config=            # URL to extra config that will be executed after installer-config.txt
     usbroot=                  # set to 1 to install to first USB disk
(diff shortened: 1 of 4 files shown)


kpfleming:
I have tested this branch with IPv4 disabled, but was unable to complete an installation in that configuration. This is not due to problems in the installer itself, but rather in various tools used during the installation (including cURL), which try to contact servers over IPv4 and don't retry when that fails. There are possible ways to address this, but I wanted to get this merged first as it represents the bulk of the work and IPv6-only networks (without NAT64/DNS64 for access to IPv4 sites) are rare.

Mausy5043:
Impressive 👍

README.md
```diff
@@ -202,8 +300,8 @@ The system is almost completely unconfigured on first boot. Here are some tasks
 The default **root** password is **raspbian**.
 
 > Set new root password: `passwd`  (can also be set during installation using **rootpw** in [installer-config.txt](#installer-customization))  
-> Configure your default locale: `dpkg-reconfigure locales`
```
Mausy5043: These two spaces force a newline.

README.md
```diff
@@ -191,8 +195,102 @@ It is possible to replace the installer script completely, without rebuilding th
...
+  Set 'ip6_nameservers' to the address of the DNS resolver which should be used (if there are multiple DNS resolvers, their addresses
+  can be included in this option, separated by spaces).
+
+- Disabled
+
+  Set 'ip6_nameservers' to 'disable'. This is only necessary to stop the installer from installing the 'rdnssd' package on networks which
+  do not provide DNS resolution over IPv6.
+
 ## Logging
-The output of the installation process is now also logged to file.
```
Mausy5043: These two spaces force a newline.

README.md
```diff
@@ -141,11 +142,14 @@ The format of the _installer-config.txt_ file and the current defaults:
                               # RPi board, your network device might be named differently. This will result in the
                               # board having no network connectivity.
     ifname=eth0
-    ip_addr=dhcp
-    ip_netmask=0.0.0.0
-    ip_broadcast=0.0.0.0
-    ip_gateway=0.0.0.0
-    ip_nameservers=
+    ip4_addr=dhcp # options are 'disable', 'dhcp', or an IPv4 address
+    ip4_prefixlength=0
```
Mausy5043: Forgive my ignorance, but why did you change `netmask` to `prefixlength`?
I thought `netmask` is a better known term.

scripts/etc/init.d/rcS
```diff
@@ -27,11 +27,14 @@ locales=
 system_default_locale=
 disable_predictable_nin=1
 ifname=eth0
-ip_addr=dhcp
-ip_netmask=0.0.0.0
-ip_broadcast=0.0.0.0
```
Mausy5043: Why remove the broadcast IP?

README.md
Mausy5043: Oh I see now. You switched to using `ip` in stead of `ifconfig`. 👍

scripts/etc/init.d/rcS
```diff
@@ -848,26 +1014,62 @@ echo $hostname > /rootfs/etc/hostname || fail
 echo "OK"
 
 echo -n "  Configuring hosts... "
-touch /rootfs/etc/hosts
-# Add localhost to hosts (if needed)
+rm /rootfs/etc/hosts
+# Add localhost to hosts
 if ! grep -q "localhost" /rootfs/etc/hosts; then
```
Mausy5043: If you `rm /rootfs/etc/hosts` (L1017) then this `grep` is not needed, me thinks.

scripts/etc/init.d/rcS
```diff
@@ -848,26 +1014,62 @@ echo $hostname > /rootfs/etc/hostname || fail
...
     echo -n "adding localhost... "
     echo "127.0.0.1 localhost" >> /rootfs/etc/hosts || fail
+    if [ -n "$have_ip6" ]; then
+        echo "::1 localhost" >> /rootfs/etc/hosts || fail
+    fi
 fi
 # Remove any existing 127.0.1.1 entries
 sed -i 's/^.*127\.0\.1\.1.*$//' /rootfs/etc/hosts
```
Mausy5043: `hosts` was `rm`-ed ...

scripts/etc/init.d/rcS
Mausy5043: obviously also due to use of `ip` I guess. ;-)

kpfleming: Yes, and also it's incredibly rare for any network to have a non-default broadcast IP address.

scripts/etc/init.d/rcS
kpfleming: Agreed, that was the last area of changes I made before submitting and clearly I didn't pay enough attention. Fixed.

scripts/etc/init.d/rcS
kpfleming: Fixed.

Mausy5043: how? I don't see the change.

scripts/etc/init.d/rcS
Mausy5043: Nothing changed?

scripts/etc/init.d/rcS
Mausy5043: 👍

kpfleming:
Ahh yes, it's amazing what happens when you modify two files but only ```git add``` one of them :-)

Mausy5043:
@diederikdehaas PR is good to go as far as I'm concerned.

Mausy5043:
@diederikdehaas bump

diederikdehaas:
I plan to (finally) work on this the coming week.

diederikdehaas:
I finally got around to testing this, sorry for that :-(

The `apt-get update` step near the end of the installation fails as dns doesn't seem to be working and as a consequence so does the installation of the firmware and kernel. 
Is that expected?
Once the system is rebooted, dns does work though.

diederikdehaas:
The reason this PR got assigned to me was because I was triggered by something you said somewhere else and that was that for a certain feature some big changes were needed in the installer. 
(as you can probably tell, I can't find that line atm)

What were those changes you had in mind?

kpfleming:
I'm fairly certain this *is* the feature that I was referring to, and the other changes were the additions of some IPv6-specific tools to the installer (which have now been incorporated into this PR instead of the separate PRs they were in previously).

kpfleming:
What was the network configuration in your test where DNS was failing? IPv4 only, IPv4+6, DHCP, ?

diederikdehaas:
IPv4 only, DHCP

diederikdehaas:
The issue is only visible by the `FAILED`, but if you apply the following patch, you'll get full details:
```diff
diederik@bagend:~/dev/raspberry-pi/raspbian-ua-netinst$ git diff
diff --git a/scripts/etc/init.d/rcS b/scripts/etc/init.d/rcS
index 8f82c20..96d0488 100755
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -1253,13 +1253,13 @@ cd $old_dir || fail
 
 
 echo -n "Updating package lists... "
-chroot /rootfs /usr/bin/apt-get update &>/dev/null
+chroot /rootfs /usr/bin/apt-get update 
 echo "OK"
 
 # firmware package can't be installed during cdebootstrap phase, so do so now
 if [ "$kernel_module" = true ] ; then
   echo -n "Installing bootloader package (=firmware)... "
-  chroot /rootfs /usr/bin/apt-get -y install raspberrypi-bootloader-nokernel &>/dev/null
+  chroot /rootfs /usr/bin/apt-get -y install raspberrypi-bootloader-nokernel 
   if [ $? -eq 0 ]; then
     echo "OK"
   else
@@ -1269,7 +1269,7 @@ fi
 
 # install libraspberrypi-bin to provide vcgencmd functionality
 echo -n "Installing libraspberrypi-bin package (=vcgencmd,raspistill,raspivid,etc)... "
-chroot /rootfs /usr/bin/apt-get -y install libraspberrypi-bin &>/dev/null
+chroot /rootfs /usr/bin/apt-get -y install libraspberrypi-bin 
 if [ $? -eq 0 ]; then
     echo "OK"
 else
```

kpfleming:
Thanks for the details, will investigate soon.

diederikdehaas:
I will too. I asked first as I wasn't sure if this was an expected issue or not.

The `resolvconf` package installs a service and [cdebootstrap-helper-rc.d](https://github.com/debian-pi/raspbian-ua-netinst/blob/v1.1.x/scripts/etc/init.d/rcS#L1158) prevents it from starting.
IIRC there was a way around that, but I'll have to dig to find it again.

kpfleming:
I suspect we'll need to just copy the /etc/resolv.conf from the installer into the rootfs for use by the chroot-ed steps, then it will be overwritten by resolvconf when the system is booted the first time.

kpfleming:
OK, I'm 99% certain that is indeed the cause of this problem, although looking through ```rcS``` in the v1.1.x branch I don't actually understand how it's working now, since the installer's resolv.conf is only copied into the rootfs when DHCP Is *not* being used. When it is, the DHCP client has populated /etc/resolv.conf with information it got from the DHCP server, but that doesn't get copied into the rootfs... so the ```apt-get update``` which is run inside the rootfs should always fail.

I'll add a commit to this branch to ensure that the installer's resolver configuration is copied into the rootfs for use before the first boot.

kpfleming:
I thought I could test this using QEMU, but it turns out that's a lot harder than it should be... I don't have any more spare Pis, so I just ordered one (my Pi1 could use an upgrade to a Pi3 anyway). Hopefully next weekend I'll be able to pick this up again.

kpfleming:
Bug confirmed and resolved. I didn't notice this before because any failure of ```apt-get update``` still reports 'OK' in the log, but I should still have noticed the two package installation failures after that. Oops!

diederikdehaas:
Awesome, thanks @kpfleming 👍

diederikdehaas:
I value all contributions, but the addition of IPv6 support in a net(work) installer is probably one of my favorites \o/

(and sorry it took so long)

diederikdehaas:
> The reason this PR got assigned to me was because I was triggered by something you said somewhere else and that was that for a certain feature some big changes were needed in the installer.

Found it!
In https://github.com/debian-pi/raspbian-ua-netinst/issues/347#issuecomment-246150536 you said:

> I've submitted PR #425 to add substantial IPv6 support (although not complete, as it does not support DHCPv6, since that would require substantial additions to the installer).

If you (still) know what those additions are, feel free to share them 👍

kpfleming:
Ahh... that. Supporting DHCPv6 would require using a full-blown DHCP client in the installer (probably dhclient, the same one that is used in the installed system) instead of udhcpc. While I haven't tried it, this would likely increase the size of the installer image quite noticeably, so as a result it's probably best to wait until the community expresses a real need for DHCPv6 support.

mutability:
Counterintutively, this PR actually broke IPv6 on an install that was previously working fine (install over IPv4, afterwards IPv6 works fine).

kpfleming:
Yes, that was intentional, since we've had reports (both related to this installer, and in general) that people have *broken* IPv6 on their networks and don't know it. When the installer did not care about IPv6, the installation would work fine, but then the resulting system would exhibit odd breakage which was difficult to diagnose because the user did not even know that there was IPv6 available on their network.

I wouldn't have any objection to changing the IPv6 default in the installer to 'auto', allowing it to be used if it's available, and then this would fall automatically into the installed system as well. If a user then reports that the installation fails and it's determined that broken IPv6 networking on their LAN is the cause, they can set it to 'disable' in their installer-config.txt file and use IPv4 only. It just seemed safer to continue ignoring IPv6 completely since that was the installer's previous behavior (although not an intentional choice).

mutability:
Well, I just undo the damage in post-install.txt now, so it's not a big deal, but it's a regression from how it behaved before in my environment.

mutability:
I have a slightly unusual environment in that I end up with v4 external connectivity and use v6 internally for a VPN. So the external connectivity doesn't matter - but globally disabling v6 breaks the VPN because it prevents setting up the tunnel device.

kpfleming:
For what it's worth, the README in the v1.1.x branch does make it clear that IPv6 support will be disabled by default, although it might be worth extending that statement to indicate that it will be disabled both during the installation and in the installed system.

PR #418, opened 2016-08-08
Rpi3 support

diederikdehaas:
There are still items to be improved, but it is essentially working on my systems.
One particular point that needs to be improved is the serial console/bluetooth. It is working during install, but during normal boot after the install it is not. And we probably need to add a config option which determines what to do with the UART port as that is either serial or bluetooth, but not both.

But for the time being people can now use the installer on a Pi 3 (but do have to build it themselves for now)

So I'd say this partially fixes issue #375

PR #429, opened 2016-09-23
fix enable_uart=1 to enable the serial console

bbinet:
see https://github.com/debian-pi/raspbian-ua-netinst/pull/418#r77345654


diederikdehaas:
Thanks :+1:

PR #431, opened 2016-09-27
Find the first unused loop device.

richard-scott:
I've updated the 'buildroot.sh' script so that it automatically finds the next free loop device and does not always try and use /dev/loop0.


diederikdehaas:
Nice 👍

PR #447, opened 2017-01-07
Added new Pi2B revision number to detection list.

diederikdehaas:


goranche:
@diederikdehaas approved

PR #460, opened 2017-03-18
fix #456 tell kernel hostname

### LINKED ISSUE #456: sudo: unable to resolve host (none)
I am using the v1.1.x branch.

In my post-install.txt I have the following commands.

```
chroot /rootfs /bin/cat /dev/zero | chroot /rootfs /usr/bin/ssh-keygen -t rsa -b 4096 -q -N ""
chroot /rootfs /bin/cat /dev/zero | chroot /rootfs /usr/bin/sudo -u user1 /usr/bin/ssh-keygen -t rsa -b 4096 -q -N ""
```

Both commands create the key files I want.
.
But the second one shows a warning.

    sudo: unable to resolve host (none)

And because of this issue, the comment in the key file is wrong ("(none)" instead of the real hostname).

How can I solve this issue?

CWempe:
make defined hostname available via `hostname` command during installation

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -864,6 +864,8 @@ else
     echo -n "adding ${hostname}.${domainname}... "
     echo "127.0.1.1 ${hostname}.${domainname} ${hostname}" >> /rootfs/etc/hosts || fail
 fi
+# Tell the kernel what the hostname is supposed to be
+hostname $hostname
 echo "OK"
 
 # networking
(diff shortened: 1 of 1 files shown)


diederikdehaas:
Why isn't this called with `chroot` so it gets set in the target system?
It now looks like it's setting the `hostname` in the installer, which then in a way get propagated into the chroot (like magic as the dns resolver somehow worked before the introduction of the `resolvconf` package)

CWempe:
I don't know.
I just had an issue (#456) and @Mausy5043 told me how to solve it. 😄

kpfleming:
The ```hostname``` command is injecting information into the kernel, not the filesystem. It won't matter whether it is run in the installer-root or the installed-root.

PR #453, opened 2017-03-14
change special character in README

CWempe:
When copy-and-paste the filename `post‑install.txt` you get an error while installing, because the hyphen was written as a special character.

```
Copying /boot files in... cp: can't create '/boot/post?install.txt': Invalid argument
```

PR #462, opened 2017-03-19
change special character in README #453

CWempe:
When copy-and-paste the filename post‑install.txt you get an error while installing, because the hyphen was written as a special character.

Copying /boot files in... cp: can't create '/boot/post?install.txt': Invalid argument


CWempe:
based on the right branch this time 😄

Mausy5043:
Looks much better. Thanks.

PR #466, opened 2017-05-19
Optimized search for required packages in package list using awk

Zeerix:
Reduces runtime of update.sh from 40s to 18s (home PC) or 38s to 11s (Cloud VM).

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -181,7 +181,7 @@ allfound() {
 }
 
 filter_package_list() {
-    grep -E 'Package:|Filename:|SHA256:|^$'
+    awk -v p="${packages[*]}" 'BEGIN{ split(p, packages) } /^Package:/{ flag=0; for (i in packages) if ($2 == packages[i]) flag=1 }; flag{ print }'
 }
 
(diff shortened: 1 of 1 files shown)


Mausy5043:
What hardware did you test this on?

Zeerix:
update.sh successfully tested on a VM (Debian, unknown hardware) and Bash for Windows 10 (also Debian, i5-6500).

Installation on Raspbi 2B was just started, but seems to be ok.

Zeerix:
gawk is default on debian, so yes.
I just tried it with "awk -P" (posix mode) and mawk, and both worked.
Anyway, both split() and the for-in loop seem to be standard awk features.

diederikdehaas:
Excellent, thanks :+1:

A while ago I was (try to) learn a bit of awk and therefor also experimented with mawk, which is my current default (`update-alternatives`) and noticed quite a difference.

Mausy5043:
I'm inclined to merge. Any objections @diederikdehaas @goranche ?

PR #467, opened 2017-05-20
Combined the two initramfs files into one

Zeerix:
* Combined the two initramfs files into one
* Added log output while extracting the packages
* Some code cleanup

Since most files in the cpio are the same, except the kernel drivers, which are in a different folder, we can just merge the files into one.
This saves space on the /boot partition for future extensions.

@TODO: Test on devices with small memory (rpi1, rpi zero).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,12 +2,13 @@
 
 set -e
 
+# configuration
 KERNEL_VERSION_RPI1=4.4.0-1-rpi
 KERNEL_VERSION_RPI2=4.4.0-1-rpi2
 
-INSTALL_MODULES="kernel/fs/btrfs/btrfs.ko"
-INSTALL_MODULES="$INSTALL_MODULES kernel/drivers/scsi/sg.ko"
-INSTALL_MODULES="$INSTALL_MODULES kernel/net/ipv6/ipv6.ko"
+INSTALL_MODULES=("kernel/fs/btrfs/btrfs.ko")
+INSTALL_MODULES+=("kernel/drivers/scsi/sg.ko")
+INSTALL_MODULES+=("kernel/net/ipv6/ipv6.ko")
 
 # checks if first parameter is contained in the array passed as the second parameter
 #   use: contains_element "search_for" "${some_array[@]}" || do_if_not_found
@@ -39,11 +40,11 @@ function check_dependencies {
         # iterate over the found dependencies
         for dep in ${deps[@]}; do
             # check if the dependency is in $modules, if not, add to temp array
-            contains_element "${dep}" "${modules[@]}" || new_found[${#new_found[@]}]="${dep}"
+            contains_element "${dep}" "${modules[@]}" || new_found=("${dep}")
         done
     done
     # add the newly found dependencies to the end of the $modules array
-    modules=("${modules[@]}" "${new_found[@]}")
+    modules+=("${new_found[@]}")
     # set the global variable to the number of newly found dependencies
     new_count=${#new_found[@]}
 }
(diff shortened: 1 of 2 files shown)


Zeerix:
Tested on Raspi 2B.

build.sh
```diff
@@ -2,12 +2,13 @@
...
 set -e
 
+# configuration
 KERNEL_VERSION_RPI1=4.4.0-1-rpi
 KERNEL_VERSION_RPI2=4.4.0-1-rpi2
 
-INSTALL_MODULES="kernel/fs/btrfs/btrfs.ko"
-INSTALL_MODULES="$INSTALL_MODULES kernel/drivers/scsi/sg.ko"
-INSTALL_MODULES="$INSTALL_MODULES kernel/net/ipv6/ipv6.ko"
+INSTALL_MODULES=("kernel/fs/btrfs/btrfs.ko")
```
goranche: the use of bash arrays requires the use of "newer" versions of bash (3.1 and higher, I think)...
while I personally don't have an issue with this, I do know of cases where this will break for the user

and true, bash 3.1 isn't really that new, but still...

Zeerix: We use bash arrays in that script already.

goranche: 🤔 have to check on that

build.sh
```diff
@@ -73,24 +74,67 @@ function create_tempfile {
...
 
+	case "$1" in
+		"rpi1")
+			KERNEL_VERSION=$KERNEL_VERSION_RPI1
+			;;
+		"rpi2")
+			KERNEL_VERSION=$KERNEL_VERSION_RPI2
+			;;
+		*)
+			echo "Invalid parameter to 'create_cpio_kernel' function!"
```
diederikdehaas: should now be `add_kernel_modules` function

build.sh
diederikdehaas: See https://github.com/debian-pi/raspbian-ua-netinst/pull/236
Given the comments in that PR, we apparently already require bash 4. Although I don't think it's unreasonable, being fully compatible with 3.1+ would be even better (which this PR seems to be)

Mausy5043: Yeah we're on bash 4

build.sh
```diff
@@ -39,11 +40,11 @@ function check_dependencies {
         # iterate over the found dependencies
         for dep in ${deps[@]}; do
             # check if the dependency is in $modules, if not, add to temp array
-            contains_element "${dep}" "${modules[@]}" || new_found[${#new_found[@]}]="${dep}"
+            contains_element "${dep}" "${modules[@]}" || new_found=("${dep}")
```
Mausy5043: not `new_found+=(` ?

Zeerix: Of course!

Mausy5043:
Please resolve the conflicts.

goranche:
I'm not sure I like this 🤔 
yes, it saves space on the SD card, but it also uses up additional space in memory. the initramfs gets loaded (and decompressed, if needed) into memory, and it stays there...

I'd rather see the two separate.
When deciding between using up more space on the SD card vs. more memory space, I'll always side with leaving more free memory, as this allows for greater future flexibility.

Mausy5043:
I'm leaning towards @goranche on this one. It usually is easier to use a bigger SD-card, whereas memory is a fixed asset.
Leaving this PR open for now. But, unless a good reason to implement this is presented I'm for closure of the PR.

Zeerix:
> yes, it saves space on the SD card, but it also uses up additional space in memory. the initramfs gets loaded (and decompressed, if needed) into memory, and it stays there...

I don't see that as a problem. Even the 128MB models can easily extract that into memory. If you are worried, that you won't have enough free memory for installation, we can add a line that unlinks the unused drivers from the ramdisk.

> When deciding between using up more space on the SD card vs. more memory space, I'll always side with leaving more free memory, as this allows for greater future flexibility.

It's more data to copy during re-sizeing the SD card. If /boot gets too big, the 128MB RAM model cannot hold it.

What do you think will get bigger faster? The drivers, or the user-mode stuff (binaries, scripts, firmware files and other data)?
You have to duplicate the user-mode stuff just to have the drivers in separate cpio files.

Mausy5043:
> we can add a line that unlinks the unused drivers from the ramdisk.

Would that be a lot of work?

Zeerix:
> > we can add a line that unlinks the unused drivers from the ramdisk.

> Would that be a lot of work?

No, I just implemented that.

When I tested this initially, I called `free` before and after removing the folder, and the memory usage went down from ca. 109 MB to 95 MB. I also checked that the other folder is gone from /lib/modules.
So it seems to work.

Mausy5043:
Looks good. Thanks.

PR #470, opened 2017-05-31
Added firmware and modules for the wlan on RPi3 and ZeroW

Zeerix:
Request for comments: What needs to be changed to get pulled?

I got my hands on a RPi3 and after some tinkering, I got the installation over onboard WLAN to work.
While adding to much firmware to this installer probably wouldn't be a good idea, the Pi 3 and the Zero W seem to have the same chip (brcm43430), and supporting the official Raspberry chipsets would be reasonable.

However, the firmware package in Raspbian is too old, we need the one from raspberrypi.org. If that would be ok, I would add that as second download mirror (like this fork does: https://github.com/FooDeas/raspberrypi-ua-netinst ).

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -1273,6 +1273,17 @@ if [ "$kernel_module" = true ] ; then
   fi
 fi
 
+# install firmware for wireless chipset (RPi 3 and Zero W)
+if [ "$ifname" != "eth0" ]; then
+  echo -n "Installing firmware for Broadcom wireless network cards... "
+  chroot /rootfs /usr/bin/apt-get -y install firmware-brcm80211 &>/dev/null
+  if [ $? -eq 0 ]; then
+    echo "OK"
+  else
+    echo "FAILED !"
+  fi
+fi
+
 # install libraspberrypi-bin to provide vcgencmd functionality
 echo -n "Installing libraspberrypi-bin package (=vcgencmd,raspistill,raspivid,etc)... "
 chroot /rootfs /usr/bin/apt-get -y install libraspberrypi-bin &>/dev/null
--- a/update.sh
+++ b/update.sh
@@ -13,10 +13,8 @@ RASPBERRYPI_ARCHIVE_KEY_FILE_NAME="raspberrypi.gpg.key"
 RASPBERRYPI_ARCHIVE_KEY_URL="${RASPBERRYPI_ARCHIVE_KEY_DIRECTORY}/${RASPBERRYPI_ARCHIVE_KEY_FILE_NAME}"
 RASPBERRYPI_ARCHIVE_KEY_FINGERPRINT="CF8A1AF502A2AA2D763BAE7E82B129927FA3303E"
 
-FIRMWARE_PACKAGE=${RASPBERRYPI_ARCHIVE_KEY_DIRECTORY}/pool/main/f/firmware-nonfree/firmware-brcm80211_0.43+rpi5_all.deb
-FIRMWARE_SHA256=a8c247ba7b39b265c3fb2042bc376cb1df08323608b2b442d4bea130782f4d7c
-
-mirror=http://archive.raspbian.org/raspbian/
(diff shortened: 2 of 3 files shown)


Zeerix:
Works great with #469

diederikdehaas:
The idea behind the `drivers_to_load` parameter is that either we or the user makes a driver package/archive which will store the necessary modules needed for 'common' or a user specific drivers needed for wireless devices and that parameter will tell the installer which drivers to load in the kernel so the device becomes usable.
I haven't looked at your other PR (yet) and that may be useful as well.
I do think that a reasonable case can be made to provide the firmware/drivers for the onboard wireless device of the RPi by default, but I haven't discussed this yet with my fellow contributors (or owners as we're now referred to here on GitHub). But as you noticed, the raspbian archives have outdated kernel/firmware :-(

Mausy5043:
I would have no reservations to merge this.

update.sh
```diff
@@ -99,14 +111,14 @@ check_key() {
     echo -n "Checking key file '${KEY_FILE}'... "
 
     # check that there is only 1 public key in the key file
-    if [ ! $(gpg --homedir gnupg --keyid-format long --with-fingerprint --with-colons ${KEY_FILE} | grep ^pub: | wc -l) -eq 1 ] ; then
+    if [ "$(gpg --homedir gnupg --keyid-format long --with-fingerprint --with-colons "${KEY_FILE}" | grep ^pub: | wc -l)" -ne 1 ] ; then
```
Mausy5043: nice

update.sh
```diff
@@ -13,6 +13,9 @@ RASPBERRYPI_ARCHIVE_KEY_FILE_NAME="raspberrypi.gpg.key"
 RASPBERRYPI_ARCHIVE_KEY_URL="${RASPBERRYPI_ARCHIVE_KEY_DIRECTORY}/${RASPBERRYPI_ARCHIVE_KEY_FILE_NAME}"
 RASPBERRYPI_ARCHIVE_KEY_FINGERPRINT="CF8A1AF502A2AA2D763BAE7E82B129927FA3303E"
 
+FIRMWARE_PACKAGE=${RASPBERRYPI_ARCHIVE_KEY_DIRECTORY}/pool/main/f/firmware-nonfree/firmware-brcm80211_0.43+rpi5_all.deb
```
diederikdehaas: there is a newer one, named `firmware-brcm80211_20161130-3+rpi1_all.deb` ... but that also has the downside of being a very specific file/version. (file sizes are the same, but the SHA256 checksum differs).
Don't have a solution (yet), but wanted to note it before I forget it.

Mausy5043: I'm not familiar with this, but is it possible to make the version a wildcard?

diederikdehaas: My bad, this was from the first commit. In the second it was replaced with a general mechanism.

Mausy5043:
Please resolve conflicts that happened after merge of #482

scripts/etc/init.d/rcS
```diff
@@ -1284,6 +1284,17 @@ if [ "$kernel_module" = true ] ; then
...
+if [ "$ifname" != "eth0" ]; then
+  echo -n "Installing firmware for Broadcom wireless network cards... "
+  chroot /rootfs /usr/bin/apt-get -y install firmware-brcm80211 &>/dev/null
+  if [ $? -eq 0 ]; then
+    echo "OK"
+  else
+    echo "FAILED !"
+  fi
+fi
+
```
Mausy5043: My concern here: Please look at lines 335-346, 605-612 & 1120-1127 -- which (also) supports WiFi dongles -- and confirm that these pieces of code will work together with your change. 
I have no possibility to test this anymore :-(

Zeerix: Confirmed.
This just installs another package with firmware blobs, which are used by the driver if needed.
The other sections install the program wpa_supplicant, which is needed as well. It handles the actual wireless connections in user mode for both on-board WiFi and dongles.

Mausy5043: OK. Thanks

PR #475, opened 2017-06-11
Change to Raspbian Kernel 4.9

Mausy5043:


BenjaminHae:
Is there a way to upgrade the kernel of existing installations? (without running the installer)

diederikdehaas:
Gonna try this soon @Mausy5043 👍

Mausy5043:
:ok:

diederikdehaas:
Just tried it out and looking at the light activity it seemed to work and according to the log file it seemed to have succeeded.
But I did encounter some issues:
- for some reason the HWRNG didn't work (see logfile)
- I couldn't follow along using my serial cable, all I saw was some garbage
- After installation, it didn't come up ...

Inspecting various files on the SD card didn't show anything weird and I'd say that it should work ...
And if I now put the power on it, I see a very brief green light, but after that only the red one. And I don't see the network come up. Also tried plugging in a monitor, but I don't see anything there either.

But I don't know whether something is wrong with the system (and what that would be) or that I'm doing something wrong. Been having loads of network issues, where I still don't know whether my router or my switch is the problem (Pi is connected to switch, which in turn is connected to my router).
And also don't know why the serial console doesn't work. Could be the way to use it is different with this kernel then before. Me not having worked on this project for a while isn't helping either ...

[PR475.install.try1.log.txt](https://github.com/debian-pi/raspbian-ua-netinst/files/1140539/PR475.install.try1.log.txt)

diederikdehaas:
> I couldn't follow along using my serial cable, all I saw was some garbage

That may be caused by me not specifying the baudrate (115200) ... :P

Mausy5043:
I've seen the HRNG message too, but didn't really look into it, because there's a software back-up. I figured the module isn't loaded or not supplied bc it's now in the kernel. I really don't know.

As far as the logfile goes it looks good to me. 
Can't see why it wouldn't boot.

Mausy5043:
Just built the `kernel-4.9` branch.
Log suggests the build went fine.

[PR475.build.try.txt](https://github.com/debian-pi/raspbian-ua-netinst/files/1143117/PR475.build.try.txt)

diederikdehaas:
> Been having loads of network issues, where I still don't know whether my router or my switch is the problem (Pi is connected to switch, which in turn is connected to my router).

I think I've done about 30 tries, but it finally succeeded \o/
So the change seems fine, my LAN ... not so much :P

oupala:
Good.

Is there a lot of work (or testing) to be done on v1.1.x before it can be released?

PR #479, opened 2017-07-15
Add HWRNG kernel module back and explicitly load the module.

diederikdehaas:
It was build-in with kernel 4.4, but with 4.9 it is once again build as
a kernel module (like it was before 4.4).


diederikdehaas:
With https://github.com/debian-pi/raspbian-ua-netinst/pull/408/commits/5edcae7f36926300c2d249dc1389b6af447d6098 (PR408) the HWRNG kernel module was added to `/etc/modules`, but that seems to not be needed anymore.

Mausy5043:
Did you install the build successfully?

diederikdehaas:
Yes. Only took 3 tries :P

PR #480, opened 2017-07-19
Disable UART by default

Mausy5043:


diederikdehaas:
👍

CWempe:
This did not work.

You also need. https://github.com/debian-pi/raspbian-ua-netinst/pull/481

PR #482, opened 2017-07-19
Blacklist rtl8192cu, let mdev use 8192cu instead

Zeerix:
* Blacklisting kernel module rtl8192cu, as it does not work properly and binds to the wireless device, preventing from loading 8192cu.
* Adding cfg80211.ko to the initramfs, as it is needed for the driver 8192cu with the current kernel (also needed for RPi3 onboard WLAN, see #470)


Zeerix:
Just successfully tested this PR and #469 on a RPi3.

PR #486, opened 2017-08-10
Add docker support for building the installer

goranche:


docker/buildroot.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+
+set -e
+
+docker_image="goranche/raspbian-ua-netinst"
```
Mausy5043: As an end-user I'd prefer to build the image myself.

docker/buildroot.sh
```diff
@@ -0,0 +1,49 @@
...
+
+if [ "${1}" != "IN_DOCKER" ]; then
+	docker run --privileged -ti --rm -v $(pwd):/raspbian-ua-netinst "${docker_image}"
+	exit
+fi
+
```
Mausy5043: This part I don't understand. Still new to Docker.

docker/buildroot.sh
goranche: so build it yourself 👍 you have the dockerfile available

the repository name on docker hub would be changed later on, of course, but for now it's under my username...

docker/buildroot.sh
goranche: uhm... ok?

goranche:
added an option to allow for running local images

docker/buildroot.sh
Mausy5043: yeah. So, here you remove the image if you pass `IN_DOCKER` as an argument?

docker/buildroot.sh
```diff
@@ -0,0 +1,49 @@
...
+EOF
+
+offset=$(($(partx --show --nr 1 -g -o START $IMG)*512))
+
+# TODO: Create a more robust version, scan for a free loop device
+# since this will run in a docker container, it is quite possible
+# loop devices will be occupied!
+losetup -o ${offset} /dev/loop0 ${IMG}
+
+mkfs.vfat /dev/loop0
```
Mausy5043: I prefer using a method similar to the one proposed (and merged) in #431.

goranche: good luck with that... 😉

Mausy5043: 😕 
This is not possible?

goranche: as stated in the comment, this needs a more robust discovery method, but it's not as easy as it sounds... don't forget this is running inside a docker container, not on a full blown OS with kernel access

docker/buildroot.sh
goranche: I have no idea what you're asking

Mausy5043: What does `docker run --privileged -ti --rm -v $(pwd):/raspbian-ua-netinst "${docker_image}"` do?

goranche: it creates a container from the image `${docker_image}`, and runs it...
the container will be removed after it's done (but not the image, that's something different altogether).
the current directory will be mapped to `/raspbian-ua-netinst` inside the container (that's how the resulting files end up in the same directory you ran it from)

this is pretty standard docker stuff, so it's documentation has tons of info on this 👍 

what did you mean by `IN_DOCKER`, though? 😊

goranche:
it would of course be useful for others to try it out, read the README, if it's clear, ... 😅

it works for me 🕶

kpfleming:
Just one comment: a Docker container *is* a full-blown system with kernel access. The only differences between the view of the system inside the container vs. outside the container is the namespaces that have been created (at a minimum a filesystem namespace, but typically also network, user, and process, and possibly others). I have no idea whether the ```loop``` device is namespace-aware, but if it is then that will make this trivial, since loop0 will always be free inside the container.

goranche:
> is a full-blown system with kernel access

there are things (mostly kernel related, module parameters for one 😓, process limitations, enumerating busses, GPIO is limited, ...) you will not be able to do from user space inside a container (without additional settings on the host system, that is, which I consider off-limit)...
I've tried setting access up without using `--privileged` on the docker command, but yeah, nope... and even so, there are issues, one of them being listing loop devices.

and no, loop0 isn't always available... learned that the hard way 😅

regarding the "is a full blown system" - it can be, sure, but that kind of defeats the purpose of using containers in the first place

goranche:
but of course, this is "arguing" about definition of terms... 😉

docker/buildroot.sh
Mausy5043: well it isn't shown here but line 7 says `if [ "${1}" != "IN_DOCKER" ]; then`

So I was guessing there would be a CLI parameter.

goranche: 😅 yes, you did 😉 

the same script is run on both the host system (as a command), as well as in the container, where it does it's magic...

it basically checks if the command line parameter is given, if not (normal condition), it will run the docker container, and inside the docker container the script will be run again, but this time with the command line parameter...

as a user, you don't need to know the parameter even exists... if you do pass it, though, it will just do a build as if it were inside the container... depending on your system, it might work, but if it's not a linux system with all dependencies installed, it will not

Mausy5043: cool

Mausy5043:
The reason I'm hesitant to merge is that this change would let a user build @goranche's version of the installer. Not the `debian-pi` one. It's not that I don't trust you... but... my paranoid mind wonders how I (as a user) can assure that I'm getting the installer that I think that I'm building.

In the installer we take a lot of effort in building from sources that we trust. Verifying signatures for every package we download. But when using Docker, what guarantees do I get that the resulting installer is not malicious?

Zeerix:
> The reason I'm hesitant to merge is that this change would let a user build @goranche's version of the installer. Not the debian-pi one.

No. You are building the depian-pi version of the installer inside a Docker container. It's running the same scripts.

The Docker container is based on the image ubuntu:16.04 from the Docker registry. You have to trust the Docker people about that.

It's like starting up a virtual machine, installing Ubuntu from a downloaded ISO image and building the installer on that. How do you know that the ISO image was OK?

Mausy5043:
@Zeerix: ISO-images usually come with a signature (MD5 or better)

Mausy5043:
> No. You are building the debian-pi version of the installer inside a Docker container. It's running the same scripts.

Right. My mistake. 😅

PR #489, opened 2017-09-18
Identify Pi Zero W etc.

dinosore:
Running v1.1.x branch on a Pi Zero W, I noticed that it was not identified as such.

rcS updated in line with current version of http://elinux.org/RPi_HardwareHistory#Board_Revision_History

### DIFF
--- a/scripts/etc/init.d/rcS
+++ b/scripts/etc/init.d/rcS
@@ -184,17 +184,20 @@ case $cpu_revision in
     0011|0014)
         rpi_hardware_version="cm"
         ;;
-    900092|900093)
+    900092|900093|920093|9000c1)
         rpi_hardware_version="0"
         ;;
-    0002|0003|0004|0005|0006|0007|0008|0009|000d|000e|000f|0010|0012|0013|0015|900032)
+    0002|0003|0004|0005|0006|0007|0008|0009|000d|000e|000f|0010|0012|0013|0015|900021|900032)
         rpi_hardware_version="1"
         ;;
-    a01041|a21041|a22042)
+    A01040|a01041|a21041|a22042)
         rpi_hardware_version="2"
         ;;
-    a02082|a22082)
+    a02082|a22082|a32082)
         rpi_hardware_version="3"
+        ;;
+    a020a0)
+        rpi_hardware_version="cm3"
         ;;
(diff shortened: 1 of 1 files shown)


scripts/etc/init.d/rcS
```diff
@@ -184,18 +184,21 @@ case $cpu_revision in
...
-    900092|900093)
+    900092|900093|920093|9000c1)
         rpi_hardware_version="0"
         ;;
-    0002|0003|0004|0005|0006|0007|0008|0009|000d|000e|000f|0010|0012|0013|0015|900032)
+    0002|0003|0004|0005|0006|0007|0008|0009|000d|000e|000f|0010|0012|0013|0015|900021|900032)
         rpi_hardware_version="1"
         ;;
-    a01041|a21041|a22042)
+    A01040|a01041|a21041|a22042)
```
Mausy5043: `A01040` ?
Or should that be `a01040`?

scripts/etc/init.d/rcS
```diff
@@ -184,18 +184,21 @@ case $cpu_revision in
...
         rpi_hardware_version="2"
         ;;
-    a02082|a22082)
+    a02082|a22082|a32082)
         rpi_hardware_version="3"
         ;;
+    a020a0)
+        rpi_hardware_version="cm3"
```
Mausy5043: There is currently no support for the CM3 hardware. So adding that here will require changes further down to specify how to treat that hardware.

scripts/etc/init.d/rcS
Mausy5043: I think should be lowercase

Mausy5043:
Thanks for your contribution. I have added a few comments.

scripts/etc/init.d/rcS
dinosore: good spot

scripts/etc/init.d/rcS
dinosore: There's already an entry for cm, and I don't think that's used either.

I'll remove cm3 if you think that's right (and cm if you want)

Mausy5043: No need. Leave it in.

PR #490, opened 2017-09-18
Don't loop forever after failure

dinosore:
This rather simple change eliminates what, to me, is the most irritating problem for blind installs.

For those who run blind installs with no intention of ever connecting a keyboard or screen, the only way to tell if something has gone wrong is to wait....................................................

Can ping the pi, but not ssh, so it must be doing something. Yes it is. It's churning out error logs or, if it goes wrong before the time has been set, the same log over and over again...

 At least if it dies, it's hinting to take out the sd card and look at the log. 

For the eagle-eyed with a screen and keyboard, there's still the option to get into a shell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,8 @@ The format of the _installer-config.txt_ file and the current defaults:
                               # "1" to enable this functionality.
     enable_uart=0             # Set to "1" to enable the UART. Disabled by default.
     gpu_mem=                  # specify the amount of RAM in MB that should be reserved for the GPU
+    try_again=0               # Specify action on failure. 0 (default) powers off. To reboot and try again,
+                              # set to 1. When set to 1, there is no limit to the number of retries.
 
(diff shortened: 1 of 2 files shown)


Mausy5043:
Thanks for your contribution. I like the idea, but wouldn't it be nice if it would try at least two or three times before giving up? 😁

dinosore:
My view is that I want to know as soon as possible that something is wrong. If the number of retries becomes configurable, I would probably set it to zero.

It all depends where it goes wrong.

If it's a typo in the config, then best to stop at once.

If it's internet connectivity, then it might be worth a retry or 3.

However, the retries might be better at a later time when connectivity may have improved.

I like the simplicity of my method.

Perhaps a compromise might be an all or nothing option.  Default not to retry with option to retry forever? 

That one's simple too - giving the option to leave it retrying overnight if it's that sort of problem.

PR #493, opened 2017-10-01
Changed kernel to 4.9.0-3

olmari:
Changed both update.sh and build.sh kernel versions to 4.9.0-3 due 4.9.0-2 exist no more in upstream.

Signed-off-by: Sami Olmari <sami@olmari.fi>

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,8 +2,8 @@
 
 set -e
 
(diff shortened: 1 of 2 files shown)


Mausy5043:
Thanks for spotting that.

PR #510, opened 2018-04-30
Change update.sh and build.sh kernel versions to 4.9.0-6

olmari:
Change update.sh and build.sh kernel versions to 4.9.0-6
due 4.9.0-4 exist no more in upstream.

Signed-off-by: Sami Olmari <sami@olmari.fi>

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -3,8 +3,8 @@
 set -e
 
 # configuration
-KERNEL_VERSION_RPI1=4.9.0-4-rpi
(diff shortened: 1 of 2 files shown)


goranche:
testing...

goranche:
works (didn't really expect it not to 😅) on a 3B+, will also try on a 2, just in case, but should be ok

goranche:
Thank you, @olmari  👍

PR #514, opened 2019-05-12
Fix almost all shellcheck issues.

diederikdehaas:
Debian Buster releases with bash 5.0 and things that worked fine with 4.4, now start giving 'weird' issues.
By fixing the issues that shellcheck (already) reported, the scripts started to work correctly consistently again.


diederikdehaas:
Turns out it has nothing to do with bash 5.0, but with one of the raspberry pi servers not being in sync with others.

